Issue #812, opened 2015-03-27
WebBrowser fails to show keypad when text field selected on android 2.3.x


codenameone:
Original [issue 812](https://code.google.com/p/codenameone/issues/detail?id=812) created by codenameone on 2013-07-30T16:45:59.000Z:

<b>Please clearly state whether the issue relates to a device/the simulator or</b>
<b>the tools.</b>
<b>If related to a device be sure to specify exactly which device with as much</b>
<b>details as possible.</b>

<b>What steps will reproduce the problem?</b>
1. Create a form using GUI Builder
2. Add a web browser component
3. Call this form from a previous form and on beforeShow, set browser url to http://www.google.com

<b>What is the expected output? What do you see instead?</b>
I expect to see and interact with the search field on google but i see google but can't type in the search field. This happens only on android 2.3.x to 4.0.x

<b>What version of the product are you using? On what operating system?</b>
Windows 8
netbeans 7.3
codenameone plugin 1.0.47
Android device Samsung s2 gingerbread 2.3.6

<b>Please provide any additional information below.</b>

I noticed that if i do a pure code form without gui builder the search field receives focus. So tried on button event generate a form and show it, but it shows the new form over the previously showing form. How do i pop old form and show only new form?

codenameone:
Closing due to inactivity.

PR #1466, opened 2015-04-26
fix for keyboard sometimes not showing up in the Android WebBrowser

Firethunder:
this will fix the issue that sometimes keyboard doesn't show up if a text input is selected.

This only occured if the WebBrowser component was initialized before it got focused. 
In this case the AndroidPeer is blocking the focusability of its descendents, which includes the WebView.

This should fix the issue #812 
And propably the problems described in the threads: https://groups.google.com/forum/?hl=en#!topic/codenameone-discussions/6bnQoaQeSgY
and 
https://groups.google.com/forum/?hl=en#!topic/codenameone-discussions/Ni628RJmTOw/aK3EjzPJ8bgJ


codenameone:
Chen, can you evaluate this?

chen-fishbein:
Seems legit, thanks

Issue #1372, opened 2015-03-27
Excessive WAKEUPS with new iOS VM


codenameone:
Original [issue 1373](https://code.google.com/p/codenameone/issues/detail?id=1373) created by codenameone on 2015-02-25T18:17:57.000Z:

The attached app runs fine, there is no issue as far as the user is concerned.

However after running the app there is a crash log on the device:

Exception Type:  EXC_RESOURCE
Exception Subtype: WAKEUPS
Exception Message: (Limit 150/sec) Observed 1081/sec over 300 secs
Triggered by Thread:  5 

In order to run this app you will need to set the property ios.fileSharingEnabled to true and you will need to copy the sample data file to the app on your device using iTunes.

Run the app and click on start. When the import completes, check the device for a crash log.

Sample program, test data file and crash log attached.

codenameone:
This maps to this issue https://code.google.com/p/codenameone/issues/detail?id=1373 which has the test case. The specific later test case is an out of memory test case and not an excessive wakeups test case. The crash I receive is a memory crash which warrants investigation but should probably be handled in a separate issue.

codenameone:
We fixed a memory leak that is probably related to the crash you are seeing. Please see if this is still happening with the current build, I'm leaving the issue open since the excessive wakeups issue is still happening.

ivanilavsky:
Hi Shai,
it's much better now. I'm not able to run the script as is. But when I use smaller SQL statements I'm able to run my 17 MB test script. 

My SQL script statements consist of 24 INSERT statements (INSERT-UNION SELECT) . It does not work even after your last code update.
But when I change my script to 6 INSERT statements in one INSERT-UNION SELECT-statement it works! It's slow but it works!
I have installations with 80 MB scripts, I will do some tests with such a big script and will post the update here if it is usable or not.

codenameone:
Hi,
I'm assuming this is triggered by a separate leak. I have a different test case that reproduces a leak that I am tracking right now.

ivanilavsky:
Hi guys, thanks for the fix, now I'm able to load my script into the database. But the performance is very poor. 17 MB script with newVM=true on iOS is running 5 minutes. (with newVM=false the same script takes only 1 minute, on Android  24 sec). Is it possible to do something with it pls? With 80MB script it is unusable. I tried all possible SQL optimizations but it simply works so slowly.

shannah:
[...]

PR #1486, opened 2015-05-08
Created native implementations for some of the heavier used String an…

shannah:
…d StringBuilder methods.  These changes were tested against https://github.com/codenameone/CodenameOne/issues/1372 .  The 15.5 meg input file now loads in 55 seconds on iPhone 4S on NewVM.  Amazingly XMLVM still runs it in 48 seconds - even though most of the hard-hitting methods are written in Java.

### DIFF
--- a/CodenameOne/src/com/codename1/impl/CodenameOneImplementation.java
+++ b/CodenameOne/src/com/codename1/impl/CodenameOneImplementation.java
@@ -63,6 +63,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Writer;
 import java.lang.ref.WeakReference;
+import java.util.ArrayList;
 
 import java.util.Enumeration;
 import java.util.Hashtable;
@@ -5933,6 +5934,33 @@ public abstract class CodenameOneImplementation {
             
             Util.cleanup(is);
             Preferences.set("cn1$InstallKey", buildKey);
+        }
+    }
+    
+    public void splitString(String source, char separator, ArrayList<String> out) {
+        int len = source.length();
+        boolean lastSeparator = false;
+        StringBuilder buf = new StringBuilder();
+        for(int iter = 0 ; iter < len ; iter++) {
+            char current = source.charAt(iter);
+            if(current == separator) {
+                if(lastSeparator) {
+                    buf.append(separator);
+                    lastSeparator = false;
+                    continue;
+                }
+                lastSeparator = true;
+                if(buf.length() > 0) {
+                    out.add(buf.toString());
+                    buf.setLength(0);
+                }
+            } else {
+                lastSeparator = false;
+                buf.append(current);
+            }
+        }
+        if(buf.length() > 0) {
+            out.add(buf.toString());
         }
     }
 }
--- a/CodenameOne/src/com/codename1/ui/Display.java
+++ b/CodenameOne/src/com/codename1/ui/Display.java
@@ -585,6 +585,7 @@ public final class Display {
             INSTANCE.dragPathX = new float[INSTANCE.PATHLENGTH];
             INSTANCE.dragPathY = new float[INSTANCE.PATHLENGTH];
             INSTANCE.dragPathTime = new long[INSTANCE.PATHLENGTH];
+            com.codename1.util.StringUtil.setImplementation(INSTANCE.impl);
             com.codename1.io.Util.setImplementation(INSTANCE.impl);
             
             // this can happen on some cases where an application was restarted etc...
--- a/CodenameOne/src/com/codename1/util/StringUtil.java
+++ b/CodenameOne/src/com/codename1/util/StringUtil.java
@@ -22,6 +22,7 @@
  */
 package com.codename1.util;
 
+import com.codename1.impl.CodenameOneImplementation;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
@@ -32,6 +33,11 @@ import java.util.Vector;
  */
 public class StringUtil {
 
+    
+    private static CodenameOneImplementation impl;
+    public static void setImplementation(CodenameOneImplementation i) {
+        impl = i;
+    }
 
     /**
      * This method replaces all occurrences of the pattern with the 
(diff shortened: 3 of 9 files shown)


shannah:
@codenameone Please review this before merging it into the master.  These changes were tested with the following test script:

~~~~
package com.codename1.issue1156;

import com.codename1.components.Progress;
import com.codename1.db.Database;
import java.io.IOException;

import com.codename1.io.Log;
import com.codename1.system.URLCallback;
import com.codename1.ui.Button;
import com.codename1.ui.Display;
import com.codename1.ui.Form;
import com.codename1.ui.events.ActionEvent;
import com.codename1.ui.events.ActionListener;
import com.codename1.ui.plaf.UIManager;
import com.codename1.ui.util.Resources;
import com.codename1.util.StringUtil;
import java.io.InputStream;
import java.util.ArrayList;
import java.util.List;

public class Issue1372Test  {

    private Form current;
    

    public void init(Object context) {
        try {
            final Resources theme = Resources.openLayered("/theme");
            UIManager.getInstance().setThemeProps(theme.getTheme(theme.getThemeResourceNames()[0]));
        } catch (final IOException e) {
            e.printStackTrace();
        }
        
        
    }

    public void start() {
        Log.p("App Arg is "+Display.getInstance().getProperty("AppArg", null));
        if (current != null) {
            current.show();
            return;
        }
        
        final Form hi = new Form("Issue 1372 Test");
        final Button startBtn = new Button("Start Test");
        
        startBtn.addActionListener(new ActionListener(){

                Display.getInstance().callSerially(new Runnable() {

                    public void run() {
                        runTest();
                    }
                    
                });
            }
            
        });
        
        
        hi.addComponent(startBtn);
        //hi.addComponent(stopBtn);
        hi.show();
        
        
        

    }

    public void stop() {
        current = Display.getInstance().getCurrent();
    }

    public void destroy() {
    }

    public void runTest(){
		
	
		final Progress dlg = new Progress("Running script ...",null);

        dlg.showModeless();
        long start = System.currentTimeMillis();
        
        Display.getInstance().invokeAndBlock(new Runnable() { 
        			
            public void run() {
           
                InputStream is=null;
				try {
					
					is = Display.getInstance().getResourceAsStream(getClass(),"/Issue1372.sql");
					executeScript(is);				
					
				} catch (IOException e) {
					e.printStackTrace();
				} catch (InterruptedException e) {
					e.printStackTrace();
				} finally {
					dlg.dispose();
					if (is!=null)
						try {
							is.close();
						} catch (IOException e) {
							e.printStackTrace();
						}
				}
            }
        });
        Log.p("Script took "+(System.currentTimeMillis()-start)+"ms");
        
	}
	
    private void executeScript(InputStream result) throws IOException, InterruptedException {

        Database db = null;

        try {

            byte[] tmp = new byte[1024];
            String previousRest = "";
            StringBuilder tmpStr = new StringBuilder(4096);

            Database.delete("tstdb");
            db = Database.openOrCreate("tstdb");
            db.beginTransaction();

            int length = result.read(tmp);
            int count = 0;
            while (length > -1)
            {
                // clear string buffer
                if (tmpStr.length() > 0) {
                    tmpStr.setLength(0);                
                }
                
                if (previousRest.trim().length() > 0) tmpStr.append(previousRest);
                
                tmpStr.append(new String(tmp, 0, length, "UTF-8"));
                
                boolean noRest = tmpStr.length()>0 && tmpStr.charAt(tmpStr.length() - 1) == '~';
                
                String str = tmpStr.toString();
                                                                    
                int tildeIdx = str.indexOf('~');
                List<String> tmpStmtsStr = StringUtil.tokenize(str, "~");
                String[] tmpStmts = tmpStmtsStr.toArray(new String[tmpStmtsStr.size()]);
                if (tmpStmts.length == 1 && tildeIdx > -1) { 
                    noRest = true;
                }

                for (int j = 0; j < tmpStmts.length; j++) {
                    if (j < tmpStmts.length - 1) {
                        String sql = tmpStmts[j].endsWith(";") ? tmpStmts[j] : tmpStmts[j] + ";";
                        db.execute(sql);                     
                    } else if (noRest) {
                        previousRest = "";
                        if (tmpStmts[j].trim().length() > 0) {
                            String sql = tmpStmts[j].endsWith(";") ? tmpStmts[j] : tmpStmts[j] + ";";
                            db.execute(sql);                           
                        }
                    } else {
                        previousRest = tmpStmts[j];
                    }
                }
                length = result.read(tmp);
                if (count++ % 100 == 0) {
                    
                    Log.p(String.valueOf(count));
                    Thread.yield();
                }
            }

            db.commitTransaction();

        } catch (IOException e) {
            db.rollbackTransaction();
            throw e;
        } catch (ArrayIndexOutOfBoundsException e) {
            db.rollbackTransaction();
            Log.e(e);
            throw e;
        } finally {
            if (db != null) db.close();
        }

    }
    
    
    

}
~~~~

The changes improve the performance from about 5 minutes to complete to 55 seconds.  I have tested on both the Old and New VM.  You can find the Issue1372.sql file [here](https://codenameone.googlecode.com/issues/attachment?aid=13730001000&name=testcase.zip&token=ABZ6GAcPIXmSoLTEUIB9A29zuz5FZus7Bg%3A1431117134238&id=1373&mod_ts_token=ABZ6GAddWxF0jHh7SZ4tsjfbI4DT3YffYQ%3A1431117134238)

codenameone:
Looks good. I think the general direction is solid.

ivanilavsky:
Perfect, thank you guys, I will try ...

ivanilavsky:
Hi guys, I'm back to work and trying to test the current version of iOS build and I found out that on my mini iPad, iOS ver. 8.3 now it does not work at all :( . The test case ends right after start without any error. The same code inside my application complains about SQL syntax and I can see that SQL statement is corrupted (beggining of statement is ok, then one part is missig and ending is from the next statement). It seems like something with charAt/tokenizer or something like that, I'm trying to localize the problem, but no success ...
I'm using still the same test sql script I posted here previously and I change my custom Utility.split to  StringUtil.tokenize as I can see in your test case here above but the result is the same.
In simulator and in Android the same code runs OK. Please help!

ivanilavsky:
Hi guys, I found out that running the sample on ipad mini iOS ver. 8.3, methods String.indexOf(), String.length(), StrignBuilder.length() return different (smaller) values like the same methods on Android and Simulator (on android and simulator are those values equal). And I think that this is the reason why StringBuilder.append() method  works incorrectly and my SQL statements are corrupted (only on iOS). May be the UTF-8 encoding is the reason  of such a behavior ...

shannah:
Thanks.  Can you provide a sample test string that exposes this?

ivanilavsky:
Hi Steve, as I'm trying to extract sample string I found out that I'm wrong. What I mentioned previously is subsequent problem. The method StringBuilder.append appends nothing, but only from one point during import in test case mentioned above: As variable count turns 114 the next iteration runs ok but only to the next command length = result.read(tmp); Bytes are probably read because length=1024 but following  tmpStr.append(new String(tmp, 0, length, "UTF-8")); appends nothing into tmpStr (the tmpStr.length() after append is 3772 on iOS, but should be 4791 as it is in Android and in simulator). Do you have any idea what is wrong?

shannah:
I posted the result of some experiments back in the original issue.  Let's pick this up there.

PR #1570, opened 2015-08-21
IOS Translator improvements (

sannysanoff:
These changes are for the local workflow with the IOS ByteCode generator.
Initially, if you ran codegen several times, it produced different results, causing XCode to reparse and recompile lots of objc dode. This happened due to the 1) timestamp modifications 2) assembler labels to depend on JVM hashcode() which is random each time.
This pull request contains following changes:
1) it offers PreservingOutputStream (drop-in replacement for FileOutputStream) which preserves timestamps if content did not change
2) for template-based files, where in-file replacements are made, it initially creates file with suffix ".new", then it replaces it to original file name, keeping timestamp if content is equal to existing one.
3) for labels, it overrides toString() adding integer sequences local to the classname, thus producting stable results each time.
 
Also pull request contains auto-generation of UIAppFonts array in -Info.plist file, because TTF file dropped into src dir did not work for me without this entry.

Also pull request contains .verbose and .draft flags of codegen taken from from System properties. "draft" flag (new) skips unused methods elimination, because for iterative development, it consumes biggest part of CPU time, and reduces generated objc linecount only by few percents (bug?)

Of course, default behaviour of flags is unchanged.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/TextArea.java
+++ b/CodenameOne/src/com/codename1/ui/TextArea.java
@@ -781,6 +781,7 @@ public class TextArea extends Component {
                 return;
             }
         }
+        System.out.println("Test5");
         Style style = getUnselectedStyle();
         rowStrings= new ArrayList();
         widthForRowCalculations = getWidth() - style.getPadding(false, RIGHT) - style.getPadding(false, LEFT);
--- a/vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeClass.java
+++ b/vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeClass.java
@@ -86,6 +86,9 @@ public class ByteCodeClass {
     }
 
     public static void markDependencies(List<ByteCodeClass> lst) {
+        if (mainClass == null) {
+            throw new RuntimeException("Missing main class..");
+        }
         mainClass.markDependent(lst);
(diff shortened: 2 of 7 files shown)


codenameone:
Thanks, but this isn't something we need. You also missed a few things which you will pay for in compilation time later on.

codenameone:
We've given some further thought to this patch and the general set of the Codename One VM and we think this would actually be an interesting enhancement. If you would like to proceed with this enhancement we can work on that. Is this patch merged with the latest version?

sannysanoff:
I added more changes in addition to those mentioned at the time of the pull request, but probably few other places remain.

But due to the nature of header file where all indexes of all methods are listed (and every .c file depends on it), referencing more new methods in java always invalidates whole project.

But still, it is big help for me, and now my project depends on my branch. At the moment I solved all tasks related to my project that required to rebuild java->ios often, but project is ongoing and I will return to this field later.

sannysanoff:
Also, unused methods optimization may be commented out as it takes 80% of compilation time and removes only 20% of the code (for my project, at least). Needs to be parametrized via cmdline, too.

codenameone:
I'd rather optimize the performance of the process rather than disable it but if you can make it into an option that would be fine.
Is your branch merged with the latest? Steve did some performance tuning work which might be valuable for your branch.

PR #1587, opened 2015-09-30
Merging Steve's fix to the tile image impl.

FabricioCabeca:


shannah:
This merge request seems to only include my iOS changes and not the rest of your changes.  It also says there are conflicts so it can't be merged directly.

I think what I'll do is fork your repo, merge the latest from the codenameone master, test it and then do a pull request from there.

PR #1608, opened 2015-11-16
fixed multipart utf8 string (fails on android)

FabricioCabeca:
There is a problem with special characters when creating a multipart request in Android at least, the string length of special characters is not the same as byte length. I've noticed it does not fail in the simulator but does fail in Android.

Ex.: addArgumentNoEncoding("city", "São Paulo");

### DIFF
--- a/CodenameOne/src/com/codename1/io/MultipartRequest.java
+++ b/CodenameOne/src/com/codename1/io/MultipartRequest.java
@@ -189,7 +189,7 @@ public class MultipartRequest extends ConnectionRequest {
                 length += baseTextLength;
                 length += key.length();
                 if(ignoreEncoding.contains(key)) {
-                    length += ((String)value).length(); 
+                    length += value.toString().getBytes().length; 
                 } else {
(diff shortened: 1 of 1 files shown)


codenameone:
Shouldn't it be `getBytes("UTF-8")`?

FabricioCabeca:
It uses the platform's default charset when not specified and worked well in my tests at least (tested with the simulator, android, iOS, my windows port). Forcing a charset may not play well with strings coming from the textfield maybe.

codenameone:
Relying on a notion of platform default in this and the `String(byte[])` constructors is the basis of 90% of encoding related issues I've run into in Java. Its never right to use that method.

The rest of the file uses UTF-8 so it seems like a sensible choice. I don't care if its configurable but using the default doesn't make sense here.

FabricioCabeca:
You are right, it could be like this instead:

```
                    try {
                        length += value.toString().getBytes("UTF-8").length;
                    } catch (UnsupportedEncodingException ex) {
                        length += value.toString().getBytes().length;
                    }

```

PR #1628, opened 2016-01-07
add event types to ActionEvent

ddyer0:
This should be harmless in the sense that existing code should be unaffected; it adds information to ActionEvent but does not alter any existing state.

The exact set of event types and names of the types is open to debate.   In a few cases I I allowed
"Other" to be the event type because a casual browse through the sources didn't reveal what would be more appropriate.


codenameone:
You see to have unresolved conflicts here. 

I suggest you checkout a fresh copy of the project merge your changes in there and commit that as it makes it a bit hard to follow the merges here. You pulled the initial copy without updating for a while (this is crucial in git) so the merges are now huge.

codenameone:
This was implemented in a separate pull request

PR #1630, opened 2016-01-08
Cumulative, reconstructed changes.

ddyer0:
Add type information for ActionEvents.  The
exact set of types and their names is subject to change, but this is a best guess.

Add printStackTrace(PrintStream), easy and useful

Make a existing drawImage variant public.

### DIFF
--- a/CodenameOne/src/com/codename1/ads/AdsService.java
+++ b/CodenameOne/src/com/codename1/ads/AdsService.java
@@ -118,7 +118,7 @@ public abstract class AdsService extends ConnectionRequest {
         if (size > 0) {
             String s = new String(out.toByteArray(), 0, size, "UTF-8");
             currentAd = s;
-            fireResponseListener(new ActionEvent(currentAd));
+            fireResponseListener(new ActionEvent(currentAd,ActionEvent.Type.Response));
         }
     }
 
--- a/CodenameOne/src/com/codename1/ads/InnerActive.java
+++ b/CodenameOne/src/com/codename1/ads/InnerActive.java
@@ -188,7 +188,7 @@ public class InnerActive extends AdsService{
 //            Storage.getInstance().flushStorageCache();            
 //            addParam(this, "cid", cid);
 //        }
-        fireResponseListener(new ActionEvent(s));        
+        fireResponseListener(new ActionEvent(s,ActionEvent.Type.Response));        
     }
 
 }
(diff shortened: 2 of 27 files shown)


codenameone:
The two other commits shouldn't be a part of this merge. The printStackTrace change won't actually work and won't be exposed to developers as a result of this change. drawImageArea doesn't work on quite a few platforms and you will run into device problems which we will then need to fix. 
Adding these sort of things to an unrelated merge makes our work harder I now can't just click the "merge pull request" button which I would have done for the main thing you submitted.

codenameone:
@shannah can you integrate only the first patch as the other two shouldn't be integrated.

shannah:
Done

ddyer0:
I can live without the public drawImageArea for now.  For the record, I would use it to
redraw a background window with a hole in it for the foreground window to remain 
untouched, by drawing up to 4 pieces that surround the hole.

The printStackTrace change is harder to dispense with.  I really like getting stack traces
in by error logs.  For which platforms is it a problem?

The current state, with one picked commit and two rejected, it seems impossible for me
to submit any new pull request without including the rejected changes.  I have a bug
fix in flowlayout.   It's a one character change - so how should I get it into your stream?

shannah:
Here is the rough process to submit only a single change.
1. Create a branch that is a copy of the CN1 master and change to that branch.
2. Cherry pick the single commit that you want to include in your pull request.
3. Create your pull request from that branch.

codenameone:
As I said above, changing these things locally won't really matter if you send a build to the build servers. 
Stuff will work on Android since it has a far more complete version of the Java runtime but on iOS this just isn't implemented. If you add additional platforms to your target (e.g. JavaScript, Windows Mobile) this might get even more problematic. So if you call these API's don't be surprised if you get build server errors when you try to target iOS etc.

ddyer0:
printStackTrace is in the VM, so presumably there would be no hope without addiing an equivalent to each vm.   I would be happy crossing those bridges one at a time.

codenameone:
Cross all of them and submit a complete patch with the C code for the iOS VM etc.
For the J2ME/RIM ports I wouldn't mind stubs that don't do anything.

ddyer0:
I don't have the capability to develop and test for all your platforms.   I can resubmit the
patch for the javase implementation, but someone else will have to add a stub to the other
vms.

PR #1632, opened 2016-01-12
Fix fencepost error in flowlayout

ddyer0:
After flow layout containers are resized to fit the contents, re-layout grew by an extra line.


ddyer0:
wrong contents.

PR #1636, opened 2016-01-13
Fix an additional error loop in the VM socket implementation

ddyer0:
If the socket is closed by the remote server, IOException was thrown without
marking the socket as unusable, which allowed to caller to keep trying.

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/AndroidAsyncView.java
+++ b/Ports/Android/src/com/codename1/impl/android/AndroidAsyncView.java
@@ -735,14 +735,16 @@ public class AndroidAsyncView extends View implements CodenameOneSurface {
              });*/
         }
 
+        private CodenameOneTextPaint font;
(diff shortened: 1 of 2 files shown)


codenameone:
Thanks!

ddyer0:
I've only patched the cases I encountered to get the desired behavior.  If you look at the broader context, there are dozens of "throw IOException" clauses that ought to be evaluated and their behavior in conjunction with higher levels tested.   Without looking, I can only guess that the other VM implementations are written in the same style.

The networking code will never be robust or reliable until it's addressed more systematically.  You're going to die the death of a thousand cuts (or bugs..) with the current code base.

codenameone:
That's unlikely. 
Sockets aren't nearly as widely used as HTTP on mobile since its just not reliable enough for most use case. 99% of our users use HTTP which we debug and use more

PR #1642, opened 2016-01-17
AndroidImplementation.java. Theme detection

McSym28:
Isn't it the right expression to detect themes on Android?

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
+++ b/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
@@ -2961,7 +2961,7 @@ public class AndroidImplementation extends CodenameOneImplementation implements 
         if (nativeThemeAvailable) {
             try {
                 InputStream is;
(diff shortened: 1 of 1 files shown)


codenameone:
I think that might introduce a compatibility problem for people relying on the old behavior. 

Is the ability to use the old theme important to your use case?

McSym28:
I'm using this ability to switch between "light" and "dark" themes in my application (it's a request from our clients).

I can change my theme handling routine to adapt to current code, but I think this can be confusing for future users.
Is it possible to introduce a new property (e.g. `"and.theme.old"`) and use it something like that:
```java
                if (android.os.Build.VERSION.SDK_INT < 14 && !isTablet() ||
                    Display.getInstance().getProperty("and.theme.old", "false").equals("true") ||
                    Display.getInstance().getProperty("and.hololight", "false").equals("true")) {
```
?

P.S. Thanks for your great work!

codenameone:
Thanks!

We can probably add that but then it would only work for you on Android instead of everywhere and would be a bit harder to customize. I think you'd be better served by using this: https://www.codenameone.com/blog/deprecations-simplified-cn1lib-installs-theme-layering.html

See the theme layering section, you can effectively toggle a theme dynamically and also see this in the simulator.

PR #1643, opened 2016-01-17
Socket.java. isSupported()

McSym28:
Doesn't this method needs to be "static"?


shai-almog:
@all-contributors please add @McSym28  for code

PR #1661, opened 2016-01-30
Master

ddyer0:
fix a problem detected by fixbugs.  As originally written, dialogTitle can only be null in the second clause.


ddyer0:
git didn't do what I expected.

PR #1688, opened 2016-02-15
Simple stack optimization & concatenated codegen

sannysanoff:
This pull requests greatly simplifies stack operations resulting assembly (e.g, i've seen ALOAD taking 2 ARM ops) via little change in method stack implementation.

Also, this pull request contains _option_ to produce concatenated .m files instead of myriads of little C files thus reducing compile time by more than 60% (on my project). (Reason of inclusion of this feature is that it was already included in my commit while cherrypicking)

### DIFF
--- a/vm/ByteCodeTranslator/src/cn1_globals.h
+++ b/vm/ByteCodeTranslator/src/cn1_globals.h
@@ -186,395 +186,395 @@ typedef struct clazz*       JAVA_CLASS;
 
 
 #define BC_ILOAD(local) { \
-    stack[stackPointer].type = CN1_TYPE_INT; \
-    stack[stackPointer].data.i = ilocals_##local##_; \
-    stackPointer++; \
+    (*SP).type = CN1_TYPE_INT; \
+    (*SP).data.i = ilocals_##local##_; \
+    SP++; \
 }
 
 #define BC_LLOAD(local) { \
-    stack[stackPointer].type = CN1_TYPE_LONG; \
-    stack[stackPointer].data.l = llocals_##local##_; \
-    stackPointer++; \
+    (*SP).type = CN1_TYPE_LONG; \
+    (*SP).data.l = llocals_##local##_; \
+    SP++; \
 }
 
 #define BC_FLOAD(local) { \
-    stack[stackPointer].type = CN1_TYPE_FLOAT; \
-    stack[stackPointer].data.f = flocals_##local##_; \
-    stackPointer++; \
+    (*SP).type = CN1_TYPE_FLOAT; \
+    (*SP).data.f = flocals_##local##_; \
+    SP++; \
 }
 
 #define BC_DLOAD(local) { \
-    stack[stackPointer].type = CN1_TYPE_DOUBLE; \
-    stack[stackPointer].data.d = dlocals_##local##_; \
-    stackPointer++; \
+    (*SP).type = CN1_TYPE_DOUBLE; \
+    (*SP).data.d = dlocals_##local##_; \
+    SP++; \
 }
 
 #define BC_ALOAD(local) { \
-    stack[stackPointer].type = CN1_TYPE_INVALID; \
-    stack[stackPointer].data.o = locals[local].data.o; \
-    stack[stackPointer].type = CN1_TYPE_OBJECT; \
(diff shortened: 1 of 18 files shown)


sannysanoff:
Benchmarks are in "UPDATE" section of document located at

https://docs.google.com/spreadsheets/u/1/d/1zALNJKvBrvfPTSe9EpsOSsSHDivmacFkIXHFwQd2MM0/pubhtml?gid=1206755131&single=true

codenameone:
Hi, 
thanks for this. It looks very interesting and we want to merge this. I have a couple of questions/comments first though:

1. Can you add license headers to the new files? Please base them on our license headers.

2. ConcatenatingFileOutputSteeam is misspelled (should be ConcatenatingFileOutputStream) but I don't quite understand what you are doing there? 
Are you generating one HUGE file of all the class files? For what purpose?

3. Is this necessary: https://github.com/sannysanoff/CodenameOne/commit/6541ae66d6ece06d02c980055af6bf3910b4aa30 ?
Based on your comment it didn't affect performance so I would guess its not. It seems like a huge change which includes risk and we'd rather avoid risk if its not essential.

4. Looking at the number columns in the chart it seems that release is performing slower than debug or am I misreading the data?

Thanks!

sannysanoff:
1) done
2) For the purpose of speeding up the compiler. It causes my project to take 2:50 instead of 6:30 in xcode to compile.
3) i am not sure which part of commit you're talking about. 
4) yes; I'm using xcode 7.2.1, I use project as it is generated, I press "debug on device" button, it uploads binary to device; then I detach debug session and launch same app in springboard. I hoped release builds will improve even more, but it didn't. I'll investigate and post more on the comment, maybe there's some unexpected xcode behavior or combination in clang flags.

sannysanoff:
Re "4)".
I observe with my own eyes (restarted xcode): Pressing "edit scheme" in xcode, seeing "Run" setting: "Debug" configuration is specified for it, but when pressing "Run", build is in progress and -Ofast is sent to compiler, although Debug configuration in project settings explicitly contains "-O0". 

Xcode sucks, because it complicates all so much. Obviously, there's a cause for this behavior, but i give up.

codenameone:
I was referring to the commits marked as "trying to approach..... nothing works yet."

Specifically changing `stack[stackPointer]` to `*SP`.

sannysanoff:
Ok, my xcode specified -O0 on the project level, but -Ofast on target level, so that was the cause. Regenerating project while it's open in xcode does not re-load that option (although normally xcode detects project changes and reloads it silently). So I'll update the spreadsheet.

----

changing stack[stackPointer] to *SP is the core of the patch. 

"stack[stackPointer + 3] = 77" translates to (unoptimized code, destination is left operand, x86-like asm)

mov reg, [stack]                        ; load variable "stack"
mov reg2, [stackPointer]           ; load variable "stackPointer"
add reg2, 3
mul reg2, sizeof struct
add reg, reg2                     ; final location
mov [reg], 77                     ; store needed value in stack

"SP[-3] = 77" translates to

mov reg, [SP]                    ; load variable
mov [reg + (3 * sizeof struct)], 77     ; store needed value in stack

so multiplication is removed completely in "SP" version.

codenameone:
Great, thanks!

codenameone:
Issue #1689 seems to be triggered by this. We also ran into some other issues and have currently reverted out servers to use the older version of the translator. I made one change trying to fix these issues but so far unsuccessfully.

JrmyDev:
I can't build anymore for iOS since 24/02/2016 (10H00 GMT+1). ... Maybe it's link with this !! It's seems the command (and '-DconcatenateFiles' parameter) is misinterpreted by the compiler :
 `java -DconcatenateFiles false ...` & `could not find or load main class false`

FYI, builds have succeeded until 23/02/2016 17H54 GMT+1 
It might be related with server/github modifications 

Thanks.

img_0001

codenameone:
That's unrelated to this issue and should be resolved now.

JrmyDev:
It's ok, thanks.

shai-almog:
@all-contributors please add @sannysanoff  for code

sannysanoff:
Cheers!

PR #1691, opened 2016-02-18
avoid an array bounds error

ddyer0:
textlength can be zero


codenameone:
The old `Math.max` code that tests for 0 is unused and should probably be removed. Text can't be 0 length because of line 820:

````java
if(text == null || text.equals("")){
       return;
}
````

ddyer0:
I actually hit the array out of bounds, so it must be because someone called setText
in another thread.  Maybe a better fix would be to use a local binding of text, so that
no matter what the value of text would be stable for the duration of the method.

PR #1702, opened 2016-02-27
Some useful error messages

PiotrZub:
No main class and no class definition error messages. Without these
message it is very hard to find out why build was failed.


PiotrZub:
No main class and no class definition error messages. Without these
message it is very hard to find out why build was failed.

PR #1707, opened 2016-03-05
Now all Location Form fields are stored and reload after next run. 

PiotrZub:
Now all Location Form fields are stored and reload after next run. Fields is updated not only after press enter but just after change value.


chen-fishbein:
Can you please update your local repository and then add the patch? at the moment git recognizes conflicts which make the merge painful

PiotrZub:
Ok. New pull request is ready.

W dniu 06-03-2016 o 16:37, Chen Fishbein pisze:

PR #1709, opened 2016-03-06
Now all values are stored and reload after next run. 

PiotrZub:
Now all values are stored and reload after next run. Fields is updated not only after press enter but just after change value.


chen-fishbein:
Thanks

PR #1729, opened 2016-04-02
Prevent build error for  Open GL ES1.

PiotrZub:


codenameone:
While this is easy to merge can you explain the motivation of using ES1?

We deprecated and no longer support the ES1 pipeline I've also noticed you complained about performance issues I'm assuming it relates to that pipeline.

PiotrZub:
Of course I do not use ES1 but it still exist in your code so I make 
this this pull request.
It was my mistake when I try do build code with USE_ES2 not defined it 
take me some time to find out why error strange build errors appears :(

Sorry, performance issues is not connected with this mistake, I test it 
with USE_ES2 defined.

shannah:
Perhaps we should just change enable the USE_ES2 constant in the source.  Currently the build server toggles this on during builds.

Re: Performance issues, please post to issue tracker and provide test case if possible.  I'll be looking into this, but it makes it much easier if you can  point me in the right direction with a test case.

codenameone:
I think this would mask a far worse issue with graphics not working correctly so this wouldn't solve the core issue rather mask it.

shai-almog:
@all-contributors please add @PiotrZub  for code

PR #1816, opened 2016-07-02
Add isLongTap method in MapComponent class to indicate if the tap on …

ahmedengu:
Add isLongTap method in MapComponent class to indicate if the tap on the map is suitable to add a map marker so this tap should be long and without dragging.

### DIFF
--- a/CodenameOne/src/com/codename1/maps/MapComponent.java
+++ b/CodenameOne/src/com/codename1/maps/MapComponent.java
@@ -1132,6 +1132,17 @@ public class MapComponent extends Container {
         }
         return super.setPropertyValue(name, value);
     }
+
+    /**
+     * Indicates whether the tap is long and without dragging.
+     *
+     * @return {@code true} if a long tap were pressed on the map without
+     * dragging, otherwise {@code false}
+     */
+    public boolean isLongTap() {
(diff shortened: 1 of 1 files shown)


codenameone:
@ahmedengu Thanks for the pull request

@chen-fishbein I'm not really familiar with that code but doesn't it make more sense to have this in the action listener callback?

ahmedengu:
Thanks for the great framework.

I was able to achieve the result of the method by working with pointer pressed listener and pointer released listener so i don't think there is a need for this pull request anymore (Sorry about that, it's the fasting effect).

But i would like to point out that getDraggedx/y() methods always return 0 when working with MapComponent which mean when calling super.pointerDragged the value of draggedx/y variables is not changing because dragAndDropInitialized is false.

An easy workaround could be adding getters to MapComponent Class for getDraggedx/y.

I'm sorry if i miss understood anything, i'm fairly new to the framework.

Best Regards,
Ahmed

codenameone:
Thanks!
The drag X/Y is related to drag and drop which is relevant when you pick up and drag a draggable component. Here you just drag the pointer so it's a more low level API so the X/Y are passed directly to the method.

PR #1819, opened 2016-07-04
"http to https" 

ahmedengu:
Greetings,

Changing http to https  to fix  "Apple will no longer accept http URL connections" 

i just changed the "http to https" for the websites that have a valid certificate only.
So i excluded : "http://tile.openstreetmap.org/, http://m2m1.inner-active.com/simpleM2M/clientRequestHtmlAd "

However you should cherry pick what is applicable to CN1.


codenameone:
Thanks!
Normally this wasn't something we would have done as most of these URL's aren't very important but it's probably a good move.

PR #1884, opened 2016-09-10
Added support for unselecting RadioButtons

ThomasH99:
Added support for unselecting RadioButtons. Useful when making ComponentGroups that allow no selection and unselecting a selected button. Should not impact existing functionality.


codenameone:
Thanks.

PR #1885, opened 2016-09-14
ToastBar.ConnectionProgress: Fix when downloading a redirected link

ahmedengu:
The toastBar was getting cleared before the file downloaded when the link is redirected 
So i moved the PROGRESS_TYPE_COMPLETED case to be a ResponseListener on the actual ConnectionRequest


codenameone:
This uses Java 8 syntax which we haven't migrated to yet in the core Codename One project we plan to do this soon at which point we will be able to accept the PR as is. If you want to resubmit with Java 5 synatx that will also work.

ahmedengu:
It's Java 5 now

codenameone:
Great, thanks!

PR #1914, opened 2016-10-16
Clean some unused variables in the Style constructor

ahmedengu:
Greetings,

There is no use for `fgSelectionColor`, `bgSelectionColor`, `scaledImage` so i cleaned them up ..
`fgSelectionColor`, `bgSelectionColor `there was no use for them that i could tell.
`scaledImage `could be used to determine scaled or tiled but tiled images have more than one type so it's kinda better to use [this constructor](https://github.com/codenameone/CodenameOne/blob/master/CodenameOne/src/com/codename1/ui/plaf/Style.java#L561)


codenameone:
Thanks!

shai-almog:
@all-contributors please add @ahmedengu  for code

PR #1969, opened 2016-11-23
[Fix] GridLayout shouldn't ignore the bottom padding

Firethunder:
GridLayout is currently ignoring the bottom padding, with this change it should work as intended


codenameone:
Ugh, thanks. That's a regression from my recent refactoring.

PR #1970, opened 2016-11-24
updated to 3.5

ThomasH99:


codenameone:
I don't understand this pull request?

codenameone:
You need to send pull requests to master never to 3.5. The pull request includes every commit we made in the past 3 months.
I'm not sure why the default table model change would be necessary, can you clarify?

Issue #1973, opened 2016-11-29
Android App crashes if Dialog.show() is called while web browser is shown


Firethunder:
Hi,
my app crashes when I try to open a dialog while the web browser is displayed .
The crash is only occurring on devices with Android 5.0.x

tested with Samsung Galaxy S4(5.0.1), Samsung Galaxy Tab A(5.0.2)

Attached is a simple test case and the produced microdump

package com.example;

import com.codename1.ui.BrowserComponent;
import com.codename1.ui.Button;
import com.codename1.ui.Dialog;
import com.codename1.ui.Display;
import com.codename1.ui.Form;
import com.codename1.ui.Toolbar;
import com.codename1.ui.events.ActionEvent;
import com.codename1.ui.events.ActionListener;
import com.codename1.ui.layouts.BorderLayout;
import com.codename1.ui.plaf.UIManager;
import com.codename1.ui.util.Resources;

/**
 * This file was generated by <a href="https://www.codenameone.com/">Codename One</a> for the purpose of building native
 * mobile applications using Java.
 */
public class App {

	private Form current;
	private Resources theme;

	public void init(Object context) {
		theme = UIManager.initFirstTheme("/theme");

		// Enable Toolbar on all Forms by default
		Toolbar.setGlobalToolbar(true);

		// Pro only feature, uncomment if you have a pro subscription
		// Log.bindCrashProtection(true);
	}

	public void start() {
		if (current != null) {
			current.show();
			return;
		}

		Form scr = new Screen();
		scr.show();

	}

	public void stop() {
		current = Display.getInstance().getCurrent();
		if (current instanceof Dialog) {
			((Dialog) current).dispose();
			current = Display.getInstance().getCurrent();
		}
	}

	public void destroy() {
	}

	private class Screen extends Form {

		public Screen() {
			setProperties();
			initBrowserComponentDemo();
			createReloadButton();
			createDialogButton();

		}

		private void setProperties() {
			setLayout(new BorderLayout());
		}

		private void initBrowserComponentDemo() {
			add(BorderLayout.CENTER, new BrowserComponentDemo());
		}

		private void createReloadButton() {
			Button reload = new Button("Reload");
			add(BorderLayout.SOUTH, reload);
			reload.addActionListener(new ActionListener<ActionEvent>() {

				@Override
				public void actionPerformed(ActionEvent evt) {
					removeAll();
					initBrowserComponentDemo();
					createReloadButton();
					createDialogButton();
					revalidate();
[...]

PR #1984, opened 2016-12-11
Fix for issue #1973

Firethunder:
This solves problem # 1973.
Android 5.0.x is having problems to remove the webview if hardware acceleration is enabled, which led to a crash. 
The solution is to simply disable the hardware acceleration when removing the webview and reenable it when restoring the webview


codenameone:
Thanks! 
Merged.

Issue #1994, opened 2016-12-23
Drop target Container doesn't scroll reasonably on dragFinished


saeder:
`com.codename1.ui.Component.dragFinishedImpl(int, int)` scrolls the drop target container using screen coordinates which often leads to strange results with scrollable containers.

Here the code to produce the misbehaviour:

	public class FormDragFinishScroll extends Form {
		public class LabelDraggable extends Label {
			public LabelDraggable(String aCaption) {
				super(aCaption);
				setDraggable(true);
				setUIID("Button");
			}
		}
	
		private class ContainerDropTarget extends Container {
			public ContainerDropTarget() {
				super(new BoxLayout(BoxLayout.Y_AXIS));
				setScrollableY(true);
				setDropTarget(true);
			}
		}
	
		public FormDragFinishScroll() {
			setTitle("FormDragFinishScroll");
			setScrollable(false);
			Container containerContent = getContentPane();
			containerContent.setLayout(new BoxLayout(BoxLayout.Y_AXIS));
			containerContent.add(new SpanLabel(
					"Drag the last entry and drop it right where it was to see the the misbehaviour"));
			ContainerDropTarget containerDropTarget = new ContainerDropTarget();
			LabelDraggable labelDraggableLast = null;
			for (int tally = 0; tally < 100; tally++) {
				LabelDraggable labelDraggable = new LabelDraggable("draggable " + (tally + 1));
				labelDraggableLast = labelDraggable;
				containerDropTarget.add(labelDraggable);
			}
			containerContent.add(containerDropTarget);
			LabelDraggable labelDraggableLast2 = labelDraggableLast;
			Display.getInstance().callSerially(() -> containerDropTarget.scrollComponentToVisible(labelDraggableLast2));
		}
	}

saeder:
This one is important because without a fix drag and drop cannot be used for scrollable components with lots of content.

saeder:
Does the call of `com.codename1.ui.Component.scrollRectToVisible (int, int, int, int, component)` in `com.codename1.ui.Component.dragFinishedImpl(int, int)` make sense at all? The draggable was already dragged there.

PR #2002, opened 2017-01-17
Scroll to the dragged component - not, as before, to absolute

saeder:
The method `com.codename1.ui.Component.dragFinishedImpl(int, int)` used to scroll to a absolute position which caused scrollable drop targets to scroll erroneously - see #1994.

This change makes it scrolling to the dragged component. What I do not know is whether this scrolling makes sense at all. But with this change it seems to work.


codenameone:
Thanks!

PR #2014, opened 2017-01-23
android peer image gets sometimes drawn incomplete

Firethunder:
this patch fixes a bug where the peer image is not drawn completly, this occurs sometimes when the paintComponent method is called shortly after a PeerComponent was removed

```
//example
removeComponent(webBrowser);
Image img = Image.createImage(webBrowser.getWidth(), webBrowser.getHeight());
webBrowser.paintComponent(img.getGraphics());
```


codenameone:
Thanks for the patch, this looks like a good direction but I'm a bit concern about potential deadlocks in this case e.g. with the UI thread waiting for the native thread.

@shannah what do you think?

shannah:
I don't think it is a good idea to use invokeAndBlock inside generatePeerImage() as this happens inside the paint method...  In addition to possible deadlocks it could fill up the invokeAndBlock queue pretty quickly. 

The use case here (Painting a native peer to a mutable image) is a bit problematic also.  When you remove the browser component (as you do in this example), that generally triggers the native peer to be either hidden or removed from the hierarchy.  Depending on the platform this may prevent it from being able to generate a peer image at all.  

For this use case, we probably need to add special support.

PR #2033, opened 2017-02-09
Added ability to perform action on ToastBar tap

diamondobama:
There are cases when ToastBar is used as an in-app notification and we want to open up another form or perform other actions.

The extra action is added to an existing leadButton and doesn't affect hiding of the ToastBar when it's touched.

Here is a simple example of usage:
`ToastBar.showMessage("You have a message from Diamond, click here to view.", FontImage.MATERIAL_CHECK_CIRCLE, 5000, e -> performAnAction());`

### DIFF
--- a/CodenameOne/src/com/codename1/components/ToastBar.java
+++ b/CodenameOne/src/com/codename1/components/ToastBar.java
@@ -50,82 +50,82 @@ import java.util.Timer;
 import java.util.TimerTask;
 
 /**
- * An API to present status messages to the user in an unobtrusive manner.  This is useful if
- * there are background tasks that need to display information to the user.  E.g. If a network request fails,
- * of let the user know that "Jobs are being synchronized".
- * 
+ * An API to present status messages to the user in an unobtrusive manner. This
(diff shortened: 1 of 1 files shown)


codenameone:
Very nice!

codenameone:
Just tried it and it seems the event isn't handled correctly probably because of the singleton nature of toolbar e.g.:

````java
        Form hi = new Form("Undo", BoxLayout.y());
        Button add = new Button("Add");
        
        add.addActionListener(e -> {
            Label l = new Label("Added this");
            hi.add(l);
            hi.revalidate();
            ToastBar.showMessage("Added, click here to undo...", FontImage.MATERIAL_UNDO, 
                    ee -> {
                        l.remove();
                        hi.revalidate();
                    });
        });
        
        hi.add(add);
        
        hi.show();
````

If you press add 3 times then press undo in the last it will remove all of the buttons meaning the action listeners of all 3 were invoked.

diamondobama:
That's right, let me try to fix that.

PR #2034, opened 2017-02-09
Fixed issue with ability to perform action on ToastBar tap

diamondobama:
I created a new leadButton and added a new ActionListener then apply it
to the current ToastBarComponent.


codenameone:
Shouldn't there be a corresponding remove listener?
With this fix there is still a leak of action listeners that could deteriorate further.

diamondobama:
Will this fix the issue, it worked fine when tested?

                c.leadButton.getListeners().clear();
                c.leadButton.addActionListener(s.getListener());
                c.leadButton.addActionListener(new ActionListener() {
                    @Override
                        if (c.currentlyShowing != null && !c.currentlyShowing.showProgressIndicator) {
                            c.currentlyShowing.clear();
                        }
                        ToastBar.this.setVisible(false);
                    }
                });

codenameone:
It relies on an implementation detail but should work for now. Thanks.

PR #2036, opened 2017-02-10
Added a constructor with uiid to Button

diamondobama:
This removes an extra line of code using a Button with custom uiid.


codenameone:
Good idea!

Issue #2012, opened 2017-01-23
Virtual Keyboard field switching too slow on older Android devices


jaanushansen:
I encountered a problem, that switching between textfields on Samsung Galaxy S III (Android 4.3) is too slow. Yes, it's a very old phone, but still it was flagship just 5 years ago.

"Slow" is very subjective claim, but I think that the issue can be improved because when 'Next' button is tapped to move from phone number entry field to PIN field for a second is shown default keyboard layout (the problem is not visible on newer phones):

![virtual_keyboard_switching](https://cloud.githubusercontent.com/assets/11716510/22201611/f4ed5e74-e16c-11e6-9301-d761e2350cc6.png)

There is also a video for the case:
https://youtu.be/0_YkraWn1pU

Test app code:

	public class MyApplication {
	
	    private Form current;
	
	    public void init(Object context) {
	    }
	    
	    public void start() {
	        if(current != null){
	            current.show();
	            return;
	        }
	        final Form hi = new Form("Welcome!", new BoxLayout(BoxLayout.Y_AXIS));
	        
	        TextField phoneNo = new TextField();
	        phoneNo.setHint("PhoneNo");
	        phoneNo.setConstraint(TextArea.PHONENUMBER);
	        hi.addComponent(phoneNo);
	        
	        TextField pin = new TextField();
	        pin.setHint("PIN");
	        pin.setConstraint(TextArea.PASSWORD | TextArea.NUMERIC);
	        hi.add(pin);        
	        hi.show();
	    }
	
	    public void stop() {
	        current = Display.getInstance().getCurrent();
	        if(current instanceof Dialog) {
	            ((Dialog)current).dispose();
	            current = Display.getInstance().getCurrent();
	        }
	    }
	    
	    public void destroy() {
	    }
	
	}

Tested on CN1 build server. I tried both "android.keyboardOpen" switches (true and false), but this config didn't seem to affect performance.

codenameone:
While we would like to address that I'm not sure we have a device to test this on.

shannah:
Had to disable this optimization by default due to https://github.com/codenameone/CodenameOne/issues/2353
It can be explicitly enabled by adding 

~~~~
Display.getInstance().getProperty("android.reuseTextEditorOnSwitch", "false")
~~~~

PR #2048, opened 2017-02-22
Example solution for issue #2012 

jaanushansen:
Example solution for issue #2012  (https://github.com/codenameone/CodenameOne/issues/2012).

Speed up switching between edited textfields on Android. Implemented and tested only in asyncEditMode. The idea is to avoid removing the native editor component just for switching time, but just native field is connected to another CN1 textfield.

But unfortunately current keyboard code is quite complex and I didn't manage to keep changes small, so probably you can't just take the changes, but at least I have given an example how the issue got improved for my test cases.


codenameone:
@shannah can you review?

shai-almog:
@all-contributors please add @jaanushansen  for code

PR #2051, opened 2017-03-02
Feature to allow null-valued properties in output from parseJSON(Reader i)

twilkinson:
This change applies to the parseJSON(Reader i) method for the JSONParser class.  As currently implemented, the Map returned by that method does not include properties from the incoming JSON stream whose values are null.  They just do not exist in the output Map.

This request implements two new methods:
public static boolean isIncludeNulls() {}
public static void setIncludeNulls(boolean aIncludeNullsDefault) {}

If set to true, properties whose values are null will be included in the Map returned by parseJSON(Reader i).

### DIFF
--- a/CodenameOne/src/com/codename1/io/JSONParser.java
+++ b/CodenameOne/src/com/codename1/io/JSONParser.java
@@ -76,6 +76,22 @@ public class JSONParser implements JSONParseCallback {
         useLongsDefault = aUseLongsDefault;
     }
 
+    /**
+     * Indicates that the parser will include null values in the parsed output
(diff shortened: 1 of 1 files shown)


codenameone:
Thanks, took me a moment to understand the use case so for other people reading it. This means that if the JSON has `x=null` within it the default behavior when I do `resultMap.keySet()` will not include `x` at all. This behavior will include it.

twilkinson:
Yes, that's correct.  Sorry, I probably could have made it clearer by including an example :-)

shai-almog:
@all-contributors please add @twilkinson  for code

PR #2054, opened 2017-03-08
fixed ability to execute tests with/without isolation between tests

JrmyDev:


JrmyDev:
Sorry this pull request contains some invalid commits

JrmyDev:
Can you merge just the commit with the fix please

codenameone:
If I understood this correctly this is a patch for one line flipping a boolean value from true to false?
Are there other changes? It's hard to see with the merges.

JrmyDev:
Yes that it. No there no other changes. Sorry I'm not so familiar with github. Actually, I just want to pull request "fix attribute cleanMode in TestRunner" commit ...

codenameone:
Github is a bit of a problem. Not sure how it got this way but the merge seems to be messed up a bit.
I'm not sure that's the right thing to do though, that flag is there to provide test isolation which is important otherwise one test can impact the result of another. Offering an option to toggle it might be OK e.g. thru a flag in codenameone_settings.properties.

JrmyDev:
I agree with you. But actually, you can't execute tests without isolation with this code !! 
As your test mecanism mixed both unit and integration tests, we may want execute some tests together. 
`TestRunner `class provide attribute for such use case but we can't switch them (I mean, if you pass `cleanMode=false` as parameter of `TestRunner `class, in build.xml `-do-test-run` task, it does nothing)

codenameone:
So the correct fix would be to fix that, no?

JrmyDev:
Absolutely. I think there is two solutions : 
- retrieve a flag in codenameone_settings.properties file (I don't know how to retrieve it in code);
- add an argument in -do-test-run task (build.xml file) (but I can't edit it and if we switch the flag, it could be erase by your update).

JrmyDev:
After reflexion, I think it should be better to accept my commit, and edit your` build.xml `file to add `cleanMode `argument by default. That way, we could be able to execute without `cleanMode `(removing arg in `build.xml` file).

codenameone:
That would take some work. Changing the build.xml means going thru a plugin update cycle and doing it for the 3 IDE's. Not trivial.

shai-almog:
@all-contributors please add @JrmyDev  for code

PR #2063, opened 2017-03-16
fixed ability to execute tests with/without isolation between tests

JrmyDev:
A proposition to fix https://github.com/codenameone/CodenameOne/pull/2054


codenameone:
Won't someone who uses `-cleanMode` reach the line with the exit error?
For compatibility I think we need to keep that flag too.

JrmyDev:
Sure. But cleanMode argument has never been useful since it's enabled by default. They just have to remove this argument.

codenameone:
I don't want to ask people who might depend on that behavior to go fix scripts. Adding a check for that flag and ignoring it is trivial, why not do it?

PR #2097, opened 2017-04-23
Optimizations for bytecode translator memory usage

ddyer0:
rewrite replaceInFile to reduce memory footprint.


vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeTranslator.java
```diff
@@ -476,18 +476,54 @@ private static String getFileType(String s) {
...
+    	// these operations efficiently.
+        for(int iter = 0 ; iter < values.length ; iter += 2) 
+        {
+        	String target = values[iter];
+        	String replacement = values[iter + 1];
+           	int found = 0;
+           	int index = 0;
+           	while((index=str.indexOf(target))>=0)
+           	{	int targetSize = target.length();
+           		str.replace(index, index+targetSize,replacement);
```
shannah: This could result in an infinite loop if `target` is a substring of `replacement`.

vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeTranslator.java
```diff
@@ -476,18 +476,54 @@ private static String getFileType(String s) {
...
+           		index += replacement.length()-targetSize;
+           		found++;
+           		totchanges++;
+           	}
+           	if(found==0)
+        		{throw new Error("target string "+target+" is missing"); }
```
shannah: Why throw an error if nothing was found to replace.  The original code didn't require this.  I.e. it was fine to try to replace a string that wasn't found.

ddyer0: It can't infinite loop - this is a single replacement and the next index starts after the 
replacement size.

Throwing the error is an element of style.  If the source doesn't contain the expected target,
would you rather be informed here, or by some undefined error down the line becuase the
file you constructed lacks something you intended to put there.

shannah: For infinite loop, look at the `while` loop.  `index` is reset in each loop iteration.  E.g. think what happens if `target="int"` and `replacement="int[]"`

I don't like the change to throw an error because it changes the way the method works.  It has been used with the assumption that you can replace a string if there is a match.  E.g. replacing all occurrences of "x" with "y", and if there are no "x", then we don't need to do anything.  I'm not aware of any (other) string pattern/replace APIs that throw an error if there are no matches, so this becomes unintuitive to use.

ddyer0:
your're right, editing error on my part, its supposed to be 
                      if(index=str.indexOf(target,index))>=0)

ddyer0:
this isn't a general purpose API, it's a constructor for xcode projects.  I can't imagine that you
would deliberated have a missed target - there was something you wanted to replace it with.
If either the source file changes format unexpectedly, or one of the replacement targets acquires
a typo, general chaos will ensue.  But hey, it's your headache when one of those things happens.

ddyer0:
..and so, I"m unsure how I can edit a pull request.  The simple thing would be for you to 
install it then add the second arg to indexOf and remove the throw if you're still unconvinced.

PR #2103, opened 2017-04-26
Optimize isMethodUsed

ddyer0:
Optimize isMethodUsed, adding new interface Signature


codenameone:
@shannah can you please review this?

ddyer0:
I'll call attention to the "isMethodUsedTester" method still embedded
in the code, which I used to verify that the revised method produces 
identical results to the original method.

PR #2107, opened 2017-04-30
return from gitSomeData if close has been requested

ddyer0:
Observed on IOS that sockets which have been requested closed can still
return true from isSocketConnected forever.  It's unclear why, but in
any case if java code has requested the socket closed, we're no longer
interested in data.  This gets processes trapped in getSomeData
released.


codenameone:
Thanks!

PR #2108, opened 2017-05-03
Support for marking threads for deletion in place & label jumb changes

ddyer0:
Eliminate the explicit list of threads that
are pending garbage collection.


codenameone:
@shannah the change with the labels isn't clear to me, I'm not sure if the deletion logic is safe as it looks like dead threads are still fair game to the GC thread and might include a race condition

ddyer0:
the label change is pretty simple, just switching a tree of strings for a hash of labels.
there are several other instances of using the labels directly instead of strings in the
same file.

The gc change is a little more subtle, but the old code depended on finalization in the
same way.

In any case, I've tested both pretty extensively, but given the pervasive impact you should
do your own testing too.  There are some commented out nslogs in the gc code.

ddyer0:
also, note both of these fix real problems.  today i dad about 10 failed builds in a row until I
managed to perturb the state of labels to succeed.  The old thread code would permanantly
lose threads if more than maxprocesses were run between gc cycles.

ddyer0:
I added a commit to this which fixes the "error 35" from pthread_create problem.

ddyer0:
this has been languishing in the queue for a long time.  Just saying... these are real problems fixed for you.

codenameone:
I agree. However, the main challenge of accepting PR's like this isn't the code it's the QA process. There are too many moving parts for comfort. If you followed the blog posts we just integrated better automated tests and are working on making them run on iOS builds. This would perform device tests automatically on a commit ensuring there are no regressions. Once this is in place it will be much easier to accept PR's like this.

shannah:
@ddyer0 For the fix for "error 35" after finite number of threads. https://github.com/codenameone/CodenameOne/pull/2108/commits/5780039380b1273251fd45c8a14c77aef7df89fa

How many threads does it generally take for you to get the error?  I'm trying to construct a test to reproduce the problem.  Currently the test looks like:

~~~~
            final int numThreads = 11002;
            final int blockSize = 10;
            int pos = 0;
            while (pos < numThreads) {
                
                Log.p("Creating threads "+pos+" to "+(pos+blockSize)+"...");
                final int[] remainingThreads = new int[]{blockSize};
                for (int i=0; i<blockSize; i++) {
                    new Thread(new Runnable() {
                        public void run() {
                            synchronized (remainingThreads) {
                                remainingThreads[0]--;
                            }
                        }
                    }).start();
                }
                while (remainingThreads[0] > 0) {
                    Display.getInstance().invokeAndBlock(()->{
                        Util.sleep(100);
                    });
                }
                pos += blockSize;
                Log.p("Finished with threads "+(pos-blockSize)+" to "+pos);
            }
~~~~

Running this test on the simulator, I'm so far unable to reproduce the error.  Haven't tried on device yet.

PR #2153, opened 2017-07-08
Updated JavaDoc

vprise:
Clarified that getCursor applies to mouse cursor to avoid confusion


shai-almog:
@all-contributors please add @vprise  for docs

PR #2182, opened 2017-08-15
150817 adding on request permission result

DurankGts:
1-New method added in LyfeCycleListener and AndroidNativeUtils to check the user permission status to android Api>=23


codenameone:
My one concern with this is that it will break compilation for anyone who uses this interface which might mean cn1libs will break. Can we have an additional interface and check whether that is implemented using `instanceof` ?

DurankGts:
Ok. Please Do that

codenameone:
I meant that you would update that

DurankGts:
I can't understand you. What do you want that I do?

codenameone:
Right now if I merge this PR we will get a compilation error for anyone who used this feature before as you added a method to an existing interface.

You need to split that to a separate interface for compatibility.

DurankGts:
Can I set to default this method to avoid problem?


public interface LifecycleListener {
    public void onCreate(Bundle savedInstanceState);
    public void onResume();
    public void onPause();
    public void onDestroy();
    public void onSaveInstanceState(Bundle b);
    public void onLowMemory();
    **public default void onRequestPermissionsResult(int requestCode,boolean permissionStatus){};**
}

codenameone:
No. We can't use Java 8 in the repos yet. We would like to add that in the future but our last attempt to do that failed

DurankGts:
do you add Java 8 to your Repository in this pull request?


No. We can't use Java 8 in the repos yet. We would like to add that in the future but our last attempt to do that failed
--

codenameone:
No. It will take a while to add that capability.

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

codenameone:
We already tried to move to Java 8 and failed. It's a very risky task as we have a lot of legacy code around and this could break builds for everyone. Right now it's not a priority but that might change.

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

codenameone:
It's hard to priorities this as it's not an urgent feature & carries a lot of risks/problems with it.

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

codenameone:
It won't be in 5.0

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

codenameone:
We don't have it scheduled yet. I suggest revising the pull request to use Java 5 syntax

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

DurankGts:
When do have programmed to include Java 8 to your Repository in this pull request?

DurankGts:
I can't understand your solution. Please provide me an example. I created this pull request because I want create my custom dialog from the runtime native permission and you have a default dialog in your code that if very ugly. this solution if to listen the result of the user when the run time native permission if showed.

codenameone:
You are adding a method to an interface. You can't do that without default methods. You will break compatibility with code that relies on that interface. You can create a separate interface e.g. `RequestPermissionResult` which implements only that method. Then:

````java
    static void onRequestPermissionsResult(int requestCode,boolean permissionStatus) {
        if(listeners != null) {
            for(LifecycleListener l : listeners) {
                if(l instanceof RequestPermissionResult) {
                    ((RequestPermissionResult)l).onRequestPermissionsResult(requestCode,permissionStatus);
                }
            }
        }
    }
````

Your concrete listener can implement both interfaces whereas everyone else will implement only the original interface.

DurankGts:
forgotten topic

PR #2193, opened 2017-08-28
280817 adding listener to accordion item

DurankGts:
adding item listener to accordion component


codenameone:
Thanks.
Your listener method is misspelled (clik instead of click). 
It doesn't comply with the convention of add/remove listener that accepts an ActionListener interface which you can then implement using EventDispatcher. This is important as EventDispatcher does quite a few things such as EDT normalization etc.

codenameone:
You also need to include JavaDoc comments for every method/class introduced.

DurankGts:
OK, please check again my last update

codenameone:
Thanks. 
I'll fix it to be addOnClick instead of setOnClick

PR #2194, opened 2017-08-28
Introduced multiple days selection on a calendar and updated Material icons to include 49 new icons from Google

diamondobama:
- Fixed selectedDay that lost its state when moving from its month to another by declaring it outside the `MonthView` class as `SELECTED_DAY`.
- Added spaces around 1-digit numbers for its width to match the 2-digits' size.
- Added `CalendarMultipleDay` style to `UIManager` for styling multiple days selection.
- Updated `FontImage` class to include a new set of icons from [Google](https://github.com/google/material-design-icons)


codenameone:
Great PR, thanks!

PR #2196, opened 2017-08-31
Calendar improvement and added actionListener to SystemTray notification.

diamondobama:


codenameone:
Thanks!
In the future can you please avoid format source for modifications? 
It just makes review of the diff much harder.

PR #2198, opened 2017-08-31
Set simulator to always show on top and other simulator improvements

diamondobama:
- Useful for while debugging on multiple screens.
- Dispose of "Please wait" Dialog if no new skin to install.
- Change skin list to radio buttons to allow easy identification of
currently used skin.


codenameone:
Great job, thanks!

PR #2201, opened 2017-09-01
010917 adding better way to custom accordion open close icons

DurankGts:
new Methods to customize Close/Open icon in Accordion component


codenameone:
Makes sense. Thanks!

PR #2204, opened 2017-09-02
Allow dates highlighting and custom uiid on `selectedDays()` method.

diamondobama:
This is a type of customization I achieved from the new Calendar class. The dots were derived from highlighting.

"20" is my selected date and "3" is the pre-selected date because it's today's date, while the rest are the highlighted dates indicating there are one or more events/plans on those dates.

Trying to mimic Google Calendar.

calendar


diamondobama:
I'm still struggling with applying a ComponentGroup to series of consequent dates I will like to group together.

codenameone:
Thanks!
FYI I noticed you used `ArrayList` in the method signatures, wouldn't it be more correct to use `Collection`?

diamondobama:
You are right! Generalization is important. My initial concern was the instantiation of the `Collection` interface, as it has many classes that implement it. I have committed a change that initializes `Collection` with an `ArrayList`.

PR #2210, opened 2017-09-05
Fixed NullPointerException and make day buttons to occupy their actual size rather than the Grid size, using FlowLayout.

diamondobama:


diamondobama:
Bear with me for several commits made to this class. I'm testing thoroughly to make it error free and more robust.

codenameone:
The flow layout enclose might be problematic, some implementations rely on the button directly in the grid. 

I think a better approach is to deprecate the usage of button and create a more elaborate `createDayComponent` method similarly to what we did in `Tree`. This will allow a developer to add an arbitrary component for day which will default to `Button` being added directly. We can have an option for a flow layout wrapper etc.

You then need to have a bind/unbind listener method that will bind the listener to the button component. That way someone can use SpanButton or a similar component and just plug it into the generic UI.

diamondobama:
Thanks for the feedback, I will do it that way.

diamondobama:
My current approach is to check if the component is a `Button`, `SpanButton` or `Container` and if it's neither, then throw an Exception. 

If it's a `Container`, set its lead component to a button and bind any `ActionListener` to this button. The current issues with this approach are `setText()` and `getText()` methods which I don't know how to handle.

diamondobama:
```
    private void bindDayListener(ActionListener l, Component cmp) {
        if (cmp instanceof Button) {
            ((Button) cmp).addActionListener(l);
        } else if (cmp instanceof SpanButton) {
            ((SpanButton) cmp).addActionListener(l);
        } else if (cmp instanceof Container) {
            if (((Container) cmp).getLeadComponent() instanceof Button) {
                ((Button) ((Container) cmp).getLeadComponent()).addActionListener(l);
            } else {
                throw new RuntimeException("Using a container requires a lead component that is a Button to function.");
            }
        } else {
            throw new RuntimeException(cmp.getClass().getSimpleName() + " is not supported. Only Button, SpanButton and Container are supported");
        }
    }
```

codenameone:
No. If you look at that stack the exception is thrown because you didn't override the bindNodeListener and setNodeIcon methods. The component can be anything.

diamondobama:
After implementation, I ended up having to override many methods for it to work. 

Is the code below normal?

```
private Calendar monthCal = new Calendar() {
    @Override
    protected Component createDayComponent() {
        Button btnDay = new Button("");
        Container cont = FlowLayout.encloseCenterMiddle(btnDay);
        cont.setLeadComponent(btnDay);
        return cont;
    }

    @Override
    protected void bindDayListener(Component cmp, ActionListener l) {
        ((Button) ((Container) cmp).getComponentAt(0)).addActionListener(l);
    }

    @Override
    protected void setDayText(Component cmp, String text) {
        ((Button) ((Container) cmp).getComponentAt(0)).setText(text);
    }

    @Override
    protected String getDayText(Component cmp) {
        return ((Button) ((Container) cmp).getComponentAt(0)).getText();
    }

    @Override
    protected void setDayUIID(Component cmp, String uiid) {
        ((Button) ((Container) cmp).getComponentAt(0)).setUIID(uiid);
    }

    @Override
    protected void setDayEnabled(Component cmp, boolean uiid) {
        ((Button) ((Container) cmp).getComponentAt(0)).setEnabled(uiid);
    }

    @Override
    protected void updateButtonDayDate(Component cmp, int currentMonth, int day) {
        if (this.isTwoDigitMode()) {
            if (day < 10) {
                ((Button) ((Container) cmp).getComponentAt(0)).setText("0" + day);
            } else {
                ((Button) ((Container) cmp).getComponentAt(0)).setText("" + day);
            }
        } else {
            if (day < 10) {
                ((Button) ((Container) cmp).getComponentAt(0)).setText(" " + day + " "); //To match the size occupied by 2 digits buttons
            } else {
                ((Button) ((Container) cmp).getComponentAt(0)).setText("" + day);
            }
        }
    }

    @Override
    protected void updateButtonDayDate(Component dayButton, int year, int currentMonth, int day) {
        updateButtonDayDate(dayButton, currentMonth, day);
    }
};
```

codenameone:
That seems normal, it's a more complex customization than we would have in Tree so I would say yes. Notice most methods are just small stubs and you don't need instanceof all over the place. You can use any arbitrary type.

BTW I would recommend using LeadComponent and not doing it the way you did it. With lead component if a touch occurs outside of the area with the button but within the cell it will still be counted.

diamondobama:
Please unmerge, I haven't committed the changes yet

PR #2215, opened 2017-09-12
JavaDoc Improvement - changed "an" to "a" where applicable

diamondobama:
Check for theme constant `hideLeftSideMenuBool` before adding a menu button.

The method below currently toggles open and close; it should do the same
for `onTopSideMenu`.

`menuButton.addActionListener(evt -> getToolbar().openSideMenu());`

### DIFF
--- a/CodenameOne/src/com/codename1/charts/models/XYSeries.java
+++ b/CodenameOne/src/com/codename1/charts/models/XYSeries.java
@@ -249,7 +249,7 @@ public class XYSeries{
   }
 
   /**
-   * Add an String at (x,y) coordinates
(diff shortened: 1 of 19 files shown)


codenameone:
Thanks!
I don't think openSidemenu should close the side menu if it's already open. That looks like a mistake.
I think you added it to workaround an issue which I already fixed with the onTopSideMenu where you see two instances one on top of the other.

diamondobama:
I reverted that part. Thought it was an intentional implementation of that method.

codenameone:
Thanks!

PR #2219, opened 2017-09-16
Thj2

ThomasH99:
Adding setAutoRelease and isAutoRelease to get SpanButton one step closer to a drop-in replacement for Button. NB Netbeans auto-reformatted the source so lots of unimportant line changes.


codenameone:
Thanks. I'm merging this since the fix is correct but in the future can you try to not format the file?
It makes reading the diff almost impossible. It also seems the commit comment is wrong.

ThomasH99:
Yes, I will avoid reformat. Sorry if the commit comment is wrong, I not well-versed in Git, can you elaborate to help me get it right next time?

codenameone:
Look above, you wrote "various updates (eg renaming getObjectId to getObjectIdP)" which seems like something you wrote for a commit in a different project?

PR #2222, opened 2017-09-21
Allow setting Calendar image in code and TextArea constructor with a uiid

diamondobama:


codenameone:
Thanks. The `TextArea` constructor might be confusing with the hint option being more common in text area/text field

diamondobama:
Reverted.

codenameone:
Thanks!

PR #2223, opened 2017-09-21
Introduced monthChangedListener that fires only when a new month is selected

diamondobama:
This is useful for loading loading events as the month changes.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Calendar.java
+++ b/CodenameOne/src/com/codename1/ui/Calendar.java
@@ -85,6 +85,7 @@ public class Calendar extends Container {
     static final long WEEK = DAY * 7;
     private EventDispatcher dispatcher = new EventDispatcher();
     private EventDispatcher dataChangedListeners = new EventDispatcher();
+    private EventDispatcher monthChangedListeners = new EventDispatcher();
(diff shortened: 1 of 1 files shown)


diamondobama:
On testing with a slow network event, the month doesn't switch until the network event is done.
Had to wrap the network event in `CN.scheduleBackgroundTask()` to allow smooth progress. 
I realize that fireActionEvent runs on EDT.

diamondobama:
Is there a better way to fire the action event without blocking the EDT?

codenameone:
The action event is on the EDT so no. I'm assuming you used addToQueueAndWait instead of addToQueue?

diamondobama:
Yes, I used `addToQueueAndWait` because I couldn't read the response data with `addToQueue`. I use `addToQueue` for background downloading only.

codenameone:
That would block the chain of events until the request (and other requests in the network queue) completes. This is problematic for something like month traversal where you might be moving fast and a request might not be relevant by the time you get the result... 

I suggest rethinking that and actually using a timer to make the request. So if after 500ms the month is still the same, make the request. Otherwise discard it. This will allow a user to page thru months without filling up the network queue.

Issue #2228, opened 2017-09-28
setHidden doesn't work correctly when scrollableX enable


JrmyDev:
After 2 hours of debugging, I finally found what cause wrong behavior when hidding a component.

Platform : Simulator 
Codename One revisions: 1b2a6d1ab1c07380ed09a71dba2f6cbd2238f2ed
Windows 7 
Eclipse Neon 4.6.3

I've extract a simple test case. It show a form with two container. The container `cnt1 `(with label 1) should be hidden, but it's the container `cnt2 `that is hidden. If you comment `cnt1.setScrollableX(true)`, no problem !

Test case : 
```
Form f = new Form();
Label lbl1 = new Label("Label1");
Label lbl2 = new Label("Label2");
Container cnt1 = FlowLayout.encloseCenter(lbl1);
cnt1.setScrollableX(true);
Container cnt2 = FlowLayout.encloseCenter(lbl2);
f.add(BorderLayout.center(cnt2).add(BorderLayout.SOUTH, cnt1));
cnt1.setHidden(true);
f.show();

```

codenameone:
I'm not sure this is a bug. `setHidden` on a scrollable container is conceptually tricky as scrollability manipulates preferred size just like `setHidden` does.. There is a simple workaround of `setScrollableX(false)` then restoring scrollability when needed.

JrmyDev:
Read again my issue please!!
First, I tried to hide cnt1 and cnt2 is hidden.
Second, don't you think you should put your workaround in your implementation ?
Finally, I think you must at least mentionned in your doc that setHidden doesn't work in case of BorderLayout/Scrollable container, even if I don't understand why you don't manage it in your implem.
I'm aware that you manage issues to avoid backward compatibility problems, but sometimes, I think you should be a bit more careful about reported problems.

codenameone:
I ran the code and saw that before posting. `setVisible` is more consistent but `setHidden` relies on nuanced behavior that's sometimes hard to quantify. I looked at the current JavaDocs and they explain what setHidden does correctly. I've added "Notice that this might not behave as expected with scrollable containers or layouts that ignore preferred size." 

I don't think that will help clarify this by much but as I said it's a nuanced behavior. Coding this workaround into our implementation doesn't make sense as it would break scrollability when you restore. If you need this behavior you need to know that you disabled scrolling so you can restore it later on. Just to be clear, this isn't about backward compatibility. It's a conceptual issue with the way preferred size works.

PR #2229, opened 2017-09-30
Fixed problem in 3 Layouts where Dimension reused between calls to getPreferredSize and getScrollSize

ThomasH99:
Border, Box and Flow layouts (contrary to all other layouts) reuse the same Dimension for calls to getPreferredSize and getScrollSize. This creates an issue since the same dim instance gets used both as preferredSize and scrollSize which creates side effects when preferredSize is forced to (0,0) in setHidden(true). (Took me several days to debug why a hidden container with BoxLayout kept becoming visible after being set hidden)

(Sorry for the Commit history, still struggling with using Git correctly)

### DIFF
--- a/CodenameOne/src/com/codename1/components/SpanButton.java
+++ b/CodenameOne/src/com/codename1/components/SpanButton.java
@@ -234,6 +234,8 @@ public class SpanButton extends Container {
      */
     public void setCommand(Command cmd) {
         actualButton.setCommand(cmd);
+        actualButton.setText(""); //remove any text set by cmd for the actualButton since text from Command should only be shown for the TextArea
+        setText(cmd.getCommandName()); //use the text of the Command (since Button is hidden)
     }
 
     /**
--- a/CodenameOne/src/com/codename1/components/SpanButton.java
+++ b/CodenameOne/src/com/codename1/components/SpanButton.java
@@ -233,8 +233,6 @@ public class SpanButton extends Container {
      * @param cmd the command
      */
     public void setCommand(Command cmd) {
-        actualButton.setCommand(cmd);
-        actualButton.setText(""); //remove any text set by cmd for the actualButton since text from Command should only be shown for the TextArea
         setText(cmd.getCommandName()); //use the text of the Command (since Button is hidden)
     }
 
--- a/CodenameOne/src/com/codename1/components/SpanButton.java
+++ b/CodenameOne/src/com/codename1/components/SpanButton.java
@@ -233,7 +233,7 @@ public class SpanButton extends Container {
      * @param cmd the command
      */
     public void setCommand(Command cmd) {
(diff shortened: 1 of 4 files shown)


codenameone:
I'm not sure if the command code in the span button is still there, it might be problematic. Moving the field into the method is probably wrong and could have a huge impact on performance. It might make sense to have two fields instead of one though.

ThomasH99:
Since there are no comments explaining the code it's difficult to me to know if there are good reasons for the current implementation. getPreferredSize gets recalculated on each call, so it shouldn't add a performance penalty. Also, since since the more recent layouts do it the same was as my change I reasoned the old layouts just never got updated. In any case, the current implementation causes the really tricky side effect I described, so something needs to be done.

codenameone:
No, preferred size is always cached. There are a lot of dependencies on this as this is pretty deep in.

ThomasH99:
I meant that it is recalculated on each call to Layout.getpreferredSize() but, I won't pretend to an expert on this But as I said, most layouts e.g. LayeredLayout, *does* create a new Dimension in each call to getPreferredSize, so I simply assumed that that was also a valid implementation and it fixes the issue. Anyway, I hope the real issue that led me to make this change was clear? If not, let me know, and I'll be happy to create an example to reproduce the issue.

codenameone:
That's a pathway that is called when layout is forced. getPreferredSize in component always caches

ThomasH99:
Here's the code to reproduce the issue. As you can see when you click the button to hide the Container it doesn't work. Hopefully this helps illustrate the issue. The reason is what I tried to describe above. 

NB. You need to click the button **several times** to see issue, first time it behaves correctly, but not after that. 
```
Form hideable = new Form("Hiding Container doesn't work", BoxLayout.y());
hideable.setLayout(new BorderLayout());
Container hideableContainer = new Container(BoxLayout.y());
hideableContainer.setScrollableY(true);
//NB. Clicking "Hide list" works the *first* time, but not after that, so click several times to see the issue
hideable.add(BorderLayout.NORTH, new Button(Command.create("Hide list",null,(e)->{
    hideableContainer.setHidden(!hideableContainer.isHidden());
    hideable.revalidate();
})));
hideable.add(BorderLayout.SOUTH, hideableContainer);
hideableContainer.addAll(new Label("label1"),new Label("label2"),new Label("label3"));
hideable.show();

```

JrmyDev:
I think it the same problem I mentionned here : https://github.com/codenameone/CodenameOne/issues/2228

ThomasH99:
Any update on this (the issue with setHidden not working is currently preventing me from building my app for a device)? Could you reproduce the issue based my example above?

codenameone:
I suggest moving that to an issue. Adding a scrollable container to the north is a mistake as the border layout always gives the full requested height to the north/south containers.

PR #2232, opened 2017-10-09
Ability to get the textarea used in SpanLabel and SpanButton

diamondobama:
Rather than having this:

        new Helper(span.getComponentAt(1)).textWhite().textCenter().ma0().pa2();

I could have below which is guaranteed to return the right component:

        new Helper(span.getTextComponent).textWhite().textCenter().ma0().pa2();


codenameone:
I guess that's just unavoidable. Thanks.

PR #2235, opened 2017-10-10
Fix for downloadImageToStorage and downloadImageToFileSystem saving error responses

diamondobama:


codenameone:
Thanks. Great catch ;-)

PR #2254, opened 2017-10-25
Added ability to create strikethrough horizontal and vertical borders

diamondobama:
I rearranged the switch cases incrementally to make them clearer before adding the strike through codes.

I couldn't include **Strike Through Horizontal** and **Strike Through Vertical** to `initComponents()` method because the code is not-editable. @codenameone, please add this for me before merging.

I had to touch `ThemeTask.java` because Underline code was not implemented and I needed to add Strike Through codes too.

Please review thoroughly before merging. I don't want to break anything.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/plaf/Border.java
+++ b/CodenameOne/src/com/codename1/ui/plaf/Border.java
@@ -68,6 +68,8 @@ public class Border {
     private static final int TYPE_OUTSET = 18;
     private static final int TYPE_IMAGE_SCALED = 19;
     private static final int TYPE_UNDERLINE = 21;
+    private static final int TYPE_STRIKETHROUGH_H = 22;
+    private static final int TYPE_STRIKETHROUGH_V = 23;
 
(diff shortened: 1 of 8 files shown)


codenameone:
Can you explain the use case of why we would need a strike through border? Notice we already have `TEXT_DECORATION_STRIKETHRU` in `Style`.

diamondobama:
Designing something like this:


strike-through-sample-h
strike-through-sample-v

codenameone:
We've got something like this in this demo: https://www.codenameone.com/demos-MSUIKit.html

Don't you think that's a bit of a niche use case? 

It isn't so much a border as a content element since it needs to align with the "ball" graphic. Keeping it in a separate layer is a recipe to problems e.g. an odd/even number of pixels might shift the border to one direction and the ball to another making them 1 pixel off.

diamondobama:
After thorough checking and evaluation, I realized that the StrikeThrough border is an overkill. The implementation in MSUIKit demo is much better.

PR #2276, opened 2017-11-12
Simple proposal to add three methods

jsfan3:
This file change corrects the javadoc of:
`public static void assertTextArea(String name, String text)`

and adds three (self explanatory) methods:
```
public static void assertTextAreaContaining(String name, String text)
public static void assertTextAreaStartingWith(String name, String text)
public static void assertTextAreaEndingWith(String name, String text)
```

I hope this can be useful in some particular circumstances.

### DIFF
--- a/CodenameOne/src/com/codename1/testing/TestUtils.java
+++ b/CodenameOne/src/com/codename1/testing/TestUtils.java
@@ -1868,7 +1868,7 @@ public class TestUtils {
         }
         TextArea l = (TextArea)findByName(name);
         assertBool(l != null, "Null area " + text);
-        assertBool(l.getText().contains(text), "assertTextArea: \"" + l.getText() + "\" is not containing: \"" + text + "\"");
+        assertBool(l.getText().indexOf(text) > -1, "assertTextArea: \"" + l.getText() + "\" is not containing: \"" + text + "\"");
 }
 
 
--- a/CodenameOne/src/com/codename1/testing/TestUtils.java
+++ b/CodenameOne/src/com/codename1/testing/TestUtils.java
@@ -1844,9 +1844,9 @@ public class TestUtils {
     }
 
     /**
-     * Asserts that we have a label with the given text baring the given name
-     * @param name the name of the label
-     * @param text the text of the label
+     * Asserts that we have a TextArea with the given text and the given name
+     * @param name the name of the TextArea
+     * @param text the text of the TextArea
      */
     public static void assertTextArea(String name, String text) {
         if(verbose) {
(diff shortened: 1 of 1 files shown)


codenameone:
This looks OK but your code doesn't compile. You relied on the contains() method of String which isn't supported by our version of String. You need to use indexOf() > -1.

jsfan3:
I replaced contains() with indexOf()>-1
What's wrong now? I see "All check have failied"...

(I'm sorry for my use of the "cointains()" method that is not available. I've previously tested the code in Netbeans and I didn't get any error. At the moment, I don't know why Netbeans confuses the Codename One API with Java8SE API in my project)

codenameone:
@shannah that looks like a bug in the CI script or is it a collision?

shannah:
That’s because I added a test for the npe in the side menu bug which hasn’t been fixed.  Tests will fail until that is fixed

codenameone:
I don't think that's a bug and this doesn't look like that.

shannah:
This is the failure.  

~~~~
[java] [EDT] 0:0:17,46 - Exception: java.lang.NullPointerException - null
     [java] java.lang.NullPointerException
     [java] 	at com.codename1.ui.SideMenuBar$CommandWrapper.actionPerformed(SideMenuBar.java:1788)
     [java] 	at com.codename1.ui.util.EventDispatcher.fireActionEvent(EventDispatcher.java:349)
~~~~

It is definitely that same bug. ((https://stackoverflow.com/questions/47141271/codename1-nullpointerexception-after-update)

Notice that the test passes up until https://github.com/codenameone/CodenameOne/commit/119120cf98796bee2a8da3f4ac758f167367b0f2, when it begins to fail.

If it's not a bug, then it's a documentation issue because the code is valid.  I can remove the test, but, just eye-balling it, I'm pretty sure I have some old code in some samples/tests from the past that would trip this as well.  Obscure, but no more obscure than the original SO poster.

In any case, I'll commit a proper fix for it tomorrow (just guarding the NPE) so that the test will pass.

In any case, the failure is not related to this pull request.

codenameone:
This shows exactly why it's a bad idea to commit failing tests. It makes the CI useless and pulls an unrelated discussion into this PR. I disagree with the premise of that test which tries to generate a failure using A deprecated API.

shannah:
I disabled the test as it was unrelated to this commit. https://github.com/shannah/cn1-unit-tests/commit/6e5bb4d0047ccac0b7891f1c22ddfab4cc81725b

This PR seems fine now.

codenameone:
Merged, thanks!

PR #2296, opened 2017-12-23
Static method to convert a Map to a valid JSON string.

jsfan3:
I noted that the CodenameOne API is missing of a method to convert a Map<String,Object> to a JSON string, that's why I created a static method to convert the given Map<String,Object> to a valid JSON representation. The values allowed types in the Map<String,Object> are: Number, String, Boolean, List, Map or null. Limited whitespace is inserted be make the resulting JSON string more readable. Simple and complex examples of usage are in the javadoc. I tested the generated JSON strings in a JSON validator. I hope that my proposed method will be helpful.

### DIFF
--- a/CodenameOne/src/com/codename1/io/JSONParser.java
+++ b/CodenameOne/src/com/codename1/io/JSONParser.java
@@ -23,7 +23,7 @@
  */
 package com.codename1.io;
 
-import com.codename1.util.StringUtil;
+import com.codename1.processing.Result;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
@@ -639,10 +639,10 @@ public class JSONParser implements JSONParseCallback {
      * Static method to convert the given {@link java.util.Map} to a valid JSON
      * representation. The values allowed types are: {@link java.lang.Number}, {@link java.lang.String}, {@link java.lang.Boolean},
      * {@link java.util.List}, {@link java.util.Map} or null.
-     * 
+     *
      * Limited whitespace is inserted be make the resulting JSON string more
      * readable.
-     * 
+     *
      * Simple example of usage:
      * <pre>{@code
      * Map<String, Object> person = new LinkedHashMap<>();
@@ -653,23 +653,19 @@ public class JSONParser implements JSONParseCallback {
      * person.put("weight (kg)", 70.7);
      *
      * Log.p("--- mapToJson() test");
-     * try {
-     *     Log.p("\n" + mapToJson(person));
-     * } catch (Exception ex) {
-     *     Log.e(ex);
-     * }
+     * Log.p("\n" + mapToJson(person));
      * }</pre>
      *
      * The output will be:
      * <pre>{@code
      * {
-     *   "firstName" : "Paco",
-     *   "lastName" : "Bellèz",
-     *   "isAlive" : true,
-     *   "age" : 35,
-     *   "weight (kg)" : 70.7
-     * }
-     * }</pre>
+     *  "firstName": "Paco",
+     *  "lastName": "Bellèz",
+     *  "isAlive": true,
+     *  "age": 35,
+     *  "weight (kg)": 70.7
+     *}
+     *}</pre>
      *
      * More complex example of usage:
      * <pre>{@code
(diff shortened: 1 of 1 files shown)


CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+     * }
+     * }</pre>
+     *
+     * @param map The map to be converted to a JSON string
+     * @return The JSON string
+     * @throws Exception A generic exception thrown when converting the map to a
+     * JSON string fails.
+     * @author Francesco Galgani
+     */
+    public static String mapToJson(Map<String, Object> map) throws Exception {
```
Maaartinus: - The argument should be `Map<String, ?>`.
- Throwing a checked exception is a big pain for such utils.

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+        int remainingItems = map.size();
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            json.append(indent).append("\"").append(entry.getKey()).append("\" : ");
+            if (entry.getValue() instanceof String) {
+                String value = (String) entry.getValue();
+                json.append("\"").append(value).append("\"");
+            } else if (entry.getValue() instanceof Number) {
+                String value;
+                if (entry.getValue() instanceof Byte) {
+                    value = ((Byte) entry.getValue()).toString();
```
Maaartinus: This cast and all following are needless. You're just testing if it's a `Number` of a kind you accept. Not sure, if it's worth it. What about `Short` or `BigInteger`?

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+                } else if (entry.getValue() instanceof Double) {
+                    value = ((Double) entry.getValue()).toString();
+                } else if (entry.getValue() instanceof Float) {
+                    value = ((Float) entry.getValue()).toString();
+                } else if (entry.getValue() instanceof Integer) {
+                    value = ((Integer) entry.getValue()).toString();
+                } else if (entry.getValue() instanceof Long) {
+                    value = ((Long) entry.getValue()).toString();
```
Maaartinus: As JS has nothing but `Double` and therefore `Long` can't be represented exactly as number, it's common to offer an option serializing `Long` to `String`.

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+                } else {
+                    throw new Exception("Cast error in converting map to json");
```
Maaartinus: Such a message isn't helpful. You need the key and the value.

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+            } else if (entry.getValue() instanceof Boolean) {
+                String value = ((Boolean) entry.getValue()).toString();
+                json.append(value);
+            } else if (entry.getValue() instanceof List) {
+                List myList = (List) (entry.getValue());
+                StringBuilder value = new StringBuilder();
+                value.append("[");
+                if (!myList.isEmpty() && (myList.get(0) instanceof String)) {
+                    for (int i = 0; i < myList.size(); i++) {
+                        String item = (String) (myList.get(i));
```
Maaartinus: - This fails badly for heterogeneous lists.
- It's needlessly repetitive and not very general, what about list of lists? It needs recursion.

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+                        Boolean item = (Boolean) (myList.get(i));
+                        value.append((Boolean) item);
+                        if (i < myList.size() - 1) {
+                            value.append(",");
+                        }
+                    }
+                } else if (!myList.isEmpty() && myList.get(0) instanceof Map) {
+                    for (int i = 0; i < myList.size(); i++) {
+                        String item = mapToJson((Map) (myList.get(i)));
+                        List<String> lines = StringUtil.tokenize(item, '\n');
```
Maaartinus: Inefficient. Use a helper method like `appendTo(StringBuilder result, Map<String, ?>, int depth)`.

And another one for lists.

CodenameOne/src/com/codename1/io/JSONParser.java
```diff
@@ -632,4 +634,270 @@ public void keyValue(String key, String value) {
...
+                        }
+                        if (i < myList.size() - 1) {
+                            value.append(",");
+                        }
+                    }
+                }
+                value.append("]");
+                json.append(value.toString());
+            } else if (entry.getValue() instanceof Map) {
+                String value = mapToJson((Map) entry.getValue());
```
Maaartinus: What if the keys aren't strings?

codenameone:
You can replace all of that code with:

````java
return Result.fromContent(map).toString();
````

It's not well documented but we use it here and there.

jsfan3:
Thank you "Maaartinus" for the very precise comments.

Thank you "codenameone" (are you Shai?) for solving the problem with one line: I tested your solution against the same Maps that I used with my code and the obtained results are equivalent.
In this new commit, I removed all my code, I inserted your one-line solution, I updated the javadoc with the example outputs resulting from Result.fromContent(map).toString();

Maybe adding this method and its javadoc can be helpfull, because otherwise it's not so intuitive how to produce and convert a Map to a Json string.

codenameone:
Thanks.
Yes I agree this method signature would help document that more effectively.

PR #2304, opened 2018-01-12
enable modification for resetScroll behavior

kutoman:
as the InfiniteContainer class does, the method has to be at least "protected" if such customizations should be allowed for own derivates

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Component.java
+++ b/CodenameOne/src/com/codename1/ui/Component.java
@@ -2543,7 +2543,7 @@ public class Component implements Animation, StyleListener {
         onScrollX(scrollX);
     }
     
-    void resetScroll() {
+    protected void resetScroll() {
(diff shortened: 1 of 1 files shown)


codenameone:
Sorry for the delay in evaluating this. This PR is a bit problematic:

- It doesn't compile since `InfiniteContainer` doesn't reflect the change
- There are no javadocs or explanation about resetScroll() that was fine for a package private undocumented method but doesn't make sense for a protected method

I don't understand why this is necessary in `InfiniteContainer` so I'm not sure I understand why it's necessary for your use case?

kutoman:
oh thanks, as you pointed at "InfiniteContainer doesn't reflect the change" I noticed that the issue I was having wasn't related to this method. My issue was that when new components were fetched my custom infinite container's scroll was reset so the first component of the view got visible. I noticed that the reset was related to another part!

PR #2323, opened 2018-02-03
Improved Validator Javadoc with an example

jsfan3:
I think that it's important to show an example of a custom validator with simple Java code (without the need of a regex), because otherwise it's not clear how to implement it.


codenameone:
Thanks!

PR #2336, opened 2018-02-14
adding callback to stream copy

kutoman:
it's my proposal to track the stream copy progress which is useful when large files needs to be copied (e.g. when downloading big files and the progress needs to be shown to the app user).

If this gets accepted, I would suggest to also extend the linked copy methods by callback parameters

### DIFF
--- a/CodenameOne/src/com/codename1/io/Util.java
+++ b/CodenameOne/src/com/codename1/io/Util.java
@@ -114,10 +114,25 @@ public class Util {
      * @param bufferSize the size of the buffer, which should be a power of 2 large enough
      */
     public static void copyNoClose(InputStream i, OutputStream o, int bufferSize) throws IOException {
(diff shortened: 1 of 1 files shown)


codenameone:
Hi,
It's missing the Consumer interface but do you know that Codename One's streams are buffered by default and observable?

https://www.codenameone.com/javadoc/com/codename1/io/BufferedInputStream.html#setProgressListener-com.codename1.io.IOProgressListener-

kutoman:
ah thanks for the hint, I didn't know that. And sorry for the Consumer interface, I defined one in my project

codenameone:
So do you think this PR is necessary in light of that functionality?  If not feel free to close it. If you think this is still something that should be used then you would need to make it compile ;-)

codenameone:
Thanks!

PR #2345, opened 2018-02-19
Change LocationManager priority will start working now

Adalbert393:
Now req parameter will be used in bindListener() and than in startUpdatingLocation. Without this commit it req parameter change nothing.

### DIFF
--- a/CodenameOne/src/com/codename1/location/LocationManager.java
+++ b/CodenameOne/src/com/codename1/location/LocationManager.java
@@ -210,8 +210,8 @@ public abstract class LocationManager {
      * to the Listener.
      */
     public void setLocationListener(final LocationListener l, LocationRequest req) {
+				request = req;
         setLocationListener(l);
-        request = req;
     }
 
     /**
--- a/CodenameOne/src/com/codename1/location/LocationManager.java
+++ b/CodenameOne/src/com/codename1/location/LocationManager.java
@@ -187,18 +187,7 @@ public abstract class LocationManager {
      * from getting updates
      */
     public void setLocationListener(final LocationListener l) {
-        synchronized (this) {
-            if (listener != null) {
-                clearListener();
-                request = null;
-                status = TEMPORARILY_UNAVAILABLE;
-            }
-            listener = l;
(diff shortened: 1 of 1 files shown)


codenameone:
I'm guessing this will collide with this code:

````java
    public void setLocationListener(final LocationListener l) {
        synchronized (this) {
            if (listener != null) {
                clearListener();
                request = null;
                status = TEMPORARILY_UNAVAILABLE;
            }
````

Adalbert393:
Yes, you're right furthermore getCurrentLocation() do not create listener so it prevent to execute clearListner(). Now I test another solution if it will be ok I create enother pull request, so you can close this pull request.

Adalbert393:
Second commit fix the problem of changing LocationManager priority

codenameone:
Thanks!

codenameone:
I've had to revert this as I got reports that basic location listener functionality stopped working after this change.

Adalbert393:
What do you mean "basic location listener functionality stopped working"? I use location listener and it is working without any problem. Maybe there is problem with proper setup LocationRequest?
For sure changing priority in Location Request is not working without this change. Every time default priority (100m accuracy) is setup.

Adalbert393:
So on simulator listener fails because an exception in StubLocationManager.java cleanListener() (Null exception) skip listener initialisation - goto catch block. But on iOS we need to cleanListener even listener and task is not started because getCurrentLocation in IOSImplementation.java use bindListener. So for now I propose check if task or timer is null in StubLocationManager.java cleanListener() becuase iOS implementation is quite complex and I am not sure if this bindListener is needed.

Adalbert393:
I think you need to open the pull request to get last commit?

codenameone:
There was a problem with set resulting in a null value after the fact. I don't see a way to re-open this PR now I think we'll need something new but I also think we need way better testing as this code seems to be very sensitive to changes of this sort. I doubt most of these changes can be covered in the unit tests of the project though.

shai-almog:
@all-contributors please add @Adalbert393  for code

Issue #2344, opened 2018-02-19
Editable vertical Slider thumb gets hidden


kutoman:
The thumb interaction does not work properly. Actually I'm not sure if it's made on purpose but the thumb never gets painted at the bottom border (see the following sample). Neither on `progress == 0` nor on `progress == 1`.
I tested on the simulator (iphone3gs)

        Form form = new Form(new BorderLayout());
        Slider vSlider = new Slider();
        vSlider.setUIID("ToastBar");
        vSlider.setVertical(true);
        vSlider.setEditable(true);
        vSlider.setThumbImage(Image.createImage(10, 30, ColorUtil.WHITE));
        vSlider.setProgress(1);
        vSlider.setMinValue(0);
        vSlider.setMaxValue(5);
        vSlider.setIncrements(1);
        
        vSlider.addDataChangedListener((type, index) -> {
            System.out.println("type: " + type + " " + index);
        });
        
        form.add(BorderLayout.WEST, vSlider);
        
        form.show();

codenameone:
Can you provide a screenshot to illustrate this?

kutoman:
or even a screen recording :)

![thumbgetsmissing](https://user-images.githubusercontent.com/5825645/36413173-8d999bde-161d-11e8-883a-43c52d53b344.gif)

PR #2348, opened 2018-02-21
fixing issue with Slider

kutoman:
for #2344

**Note:** This only fixed the issue for my use case! I don't know if it's sufficient for other Slider use cases.


codenameone:
Thanks!

PR #2417, opened 2018-05-18
Add SetProperty as a new type of property

ramsestom:
This PR should be easier to review than my previous ones ;)
It adds the SetProperty property type and adds the addAll() and removeAll() methods to ListProperty (because in some case you want to be able to add or remove a set of values to a collection property without firing a change event for each insertion or deletion (but only at the end of the process))


codenameone:
Thanks that's much easier!

PR #2418, opened 2018-05-18
180518_1137pm _custom background uiid open close icon

DurankGts:
Possibility to custom the Background Open/close icon in accordion with out use AccordionArrow uiid hardcode in line 322 of the Accordion class


codenameone:
But why use `MultiButton` and why expose `addContent(MultiButton header, Component body)` when `addContent(Component header, Component body)` is literally the same method? Why not just have a setter for the UIID?

DurankGts:
Simply. Because is more easy to custom uidd to the Multibutton out side the Accordion instance that adding a simple setter.

DurankGts:
Adding the MultiButton it give us the possibility to add various line in the Header of the Accordion, at the same time that to custom it outside of the Accordion instance. you should used Multibutton instead A Button in the Inner class AccordionContent. The main problem is to custom the uiid of the OpenClose button and its Background  in the accordion.

codenameone:
Is this PR superseded by #2424 ?
My problem with this is that you can already add a header that's a multibutton since it's a component. You are effectively exposing an implementation detail to the user. Why expose that? 
I don't understand that from the code.

PR #2424, opened 2018-05-21
210518 1006 pm custom back ground open close icon in accordion

DurankGts:
1-Possibility to custom uiid in open / close icon with setter
2-Possibility to custom uiid header with setter


codenameone:
I'll merge this but I'll make a couple of fixes most importantly the `AccordionHeader` UIID isn't defined so for compatibility I'll restore `AccordionArrow` as the default value.

PR #2430, opened 2018-05-26
Device resolution refactor

ramsestom:
This should fix the performances issue. 
By security I also implemented a cache for fontScale on android as I don't know if getContext().getResources().getConfiguration().fontScale is costly.
As a side effect, if a user change its prefered font scale in android accessibility parametters, he would now have to restart the app for this to be taken into account. But I don't think this is that much of a deal


codenameone:
I'll try to get to it. I've got a few deadlines coming so I've got low availability this week.

codenameone:
I just noticed from user contacts that the previous PR commented out `CN.getDeviceDPI()`. That's a HUGE problem...

ramsestom:
you mean `getDeviceDensity()` ? If yes just uncomment it. It was commented when I removed getDeviceDensity() in Display and other methods in the first PR and I just forgot to put it back when I restored this getDeviceDensity() method as `CN.getDeviceDensity()` is not used anywhere else in the current CN1 code...

codenameone:
You can't comment out/remove or even change the signature of a public method that's outside the impl package. That breaks binary compatibility. The new patch still has merge conflicts but it also means we'll need a deeper review of this to avoid these problems. That's generally why we prefer smaller PR's.

ramsestom:
closing this as I am planing to split the initial PR into smaller PRs

PR #2431, opened 2018-05-26
Modifications of the AnimationManager to allow parallel animations

ramsestom:
This PR modify the AnimationManager to allow running some animations in parallel when needed and achieve effects like this:
https://streamable.com/465jv

By default, animations are still run serially in a single queue so this should not break anything

The only thing I couldn't determine is whether you prefer the current flushAnimation(Runnable r) method to map to the flushAnimation(Runnable r) (that would execute the runnable when all animations of queue 0 (= the default queue) have ended) or to flushAnimationAll(Runnable r) (that would execute the runnable when all animations of all parallel queues have ended). 
It do not make any difference with the current CN1 code as CN1 only use serial animations for now (that would all be placed in queue 0) but it might make more sense to rename flushAnimationAll(Runnable r) to flushAnimation(Runnable r) and flushAnimation(Runnable r) ro flushAnimationMainQueue(Runnable r). As you whish ;)

### DIFF
--- a/CodenameOne/src/com/codename1/ui/AnimationManager.java
+++ b/CodenameOne/src/com/codename1/ui/AnimationManager.java
@@ -37,71 +37,136 @@ import java.util.ArrayList;
  */
 public final class AnimationManager {
     private final Form parentForm;
-    private ArrayList<ComponentAnimation> anims = new ArrayList<ComponentAnimation>();
-    private ArrayList<Runnable> postAnimations = new ArrayList<Runnable>();
-    
+    private ArrayList<ArrayList<ComponentAnimation>> anims_queues = new ArrayList<ArrayList<ComponentAnimation>>(); //animations queues. Animations of a same queue would be run in serie while queues run in parrallel
+    private ArrayList<ArrayList<Runnable>> postAnimations_queues = new ArrayList<ArrayList<Runnable>>(); //runnables that would run when all animations of a specific queue have finished 
+    private ArrayList<Runnable> postAllAnimations =  new ArrayList<Runnable>(); //runnables that would run when all animations, of all queues, have finished 
+        
     AnimationManager(Form parentForm) {
         this.parentForm = parentForm;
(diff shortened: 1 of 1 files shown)


codenameone:
There is already support for concurrent animations via `ComponentAnimation.compoundAnimation()`  see https://www.codenameone.com/blog/composite-animations.html

The animation manager is specifically sequential to keep the complexity of the individual animation encapsulated.

ramsestom:
I am well aware of the `compoundAnimation` and I am already using it. But you can't achieve parallel independant animations (like at 8s of the video where the folding of the field `First name' occur at the same time as the unfolding of the `Last name` field) with `compoundAnimation` .
compoundAnimation is meant to handle animations that are triggered by the same event (like a click on the `First name' field that would produce its folding/unfolding that requiere both a position and style change animations)
When you want to run two animations in parallel, each triggered by their own event (like a click on the `First name' field for one animation, and a click on the `Last name' field for the other), there is currently now way to do it with the CN1 implementation. 
Try to achieve the same effect as in the video with CN1 TextComponent fields (in slow motion) and you will see that you can't.
Anyway, I refactored a bit my code and finally took a final decision on what method  `flushAnimation(Runnable r)` should trigger . And this is the implementation I would be using in my local version of CN1 and for my own components

codenameone:
That's a great example of animations that aren't independent. 
When we launched we didn't have the `AnimationManager`, it's a recent addition. We allowed anyone to add an animation anytime and it usually worked. 

The example you gave is the perfect example of why a queue in the animation manager is essential. Both animations are mutating the state of the same container. If they would run in parallel but independently they will collide when they mutate the internal state of `Container`.  

I don't think this is crucial for fast speed animations but if it's important to you I would suggest a different route outside of animation manager. Create a special case detection system and use compound animation. Delay the original animation by a few milliseconds to see if another animation is coming in and then wrap both.

Issue #2257, opened 2017-10-28
RFE: SideMenu on right of Toolbar


jsfan3:
At the moment, the SideMenu can be added to the left of the Toolbar using addMaterialCommandToSideMenu. 
Can you add a simple way to add a SideMenu to the right of the Toolbar?

jsfan3:
This RFE can be closed.
https://www.codenameone.com/blog/right-sidemenu-tab-order.html

PR #2437, opened 2018-06-03
Added right sidemenu

jsfan3:
I did the best I could to implement a right sidemenu, as in RFE https://github.com/codenameone/CodenameOne/issues/2257 and as suggested by Shai in Stack Overflow: https://stackoverflow.com/a/50653946/2670744

I use the "format" option of Netbeans when I write code to keep it well formatted, that's why there are a lot of unnecessary changes in the `Toolbar.java`, I'm so sorry about that.

I added several methods to support the right sidemenu, such as: `addCommandToRightSideMenu, addComponentToRightSideMenu, addMaterialCommandToRightSideMenu, closeRightSideMenu, openRightSideMenu, isComponentOnTopRightSideMenu, setComponentToRightSideMenuSouth, constructOnTopRightSideMenu, constructPermamentRightSideMenu, getRightSideMenuCommands, etc.`. In this video you can see an example with the Simulator, in which I added a left sidemenu (without hamburger menu) and a right sidemenu (with the hamburger menu):
https://www.informatica-libera.net/rightSidemenu.mp4

I used as much as possible the same logic used in the left side menu bar to create a similar right side menu bar: the code is as much as possible similar.

However, in this implementation there are some limits / bugs that can require your attention:
1. The right sidemenu can be used only as a permanent sidemenu or as an onTop sidemenu, otherwise the methods that I added do nothing.
2. I didn't implemented the swipe to open and close the right sidemenu: more precisely, in the method `constructOnTopRightSideMenu` I commented the `parent.addPointerDraggedListener(new ActionListener() {...}` because the PointerDraggedListener seems to detect only the drag from the left edge to the center, and not from the right edge to the center. I'm also not sure how to manage two PointerDraggedListeners (for the left sidemenu and for the right sidemenu) and how to detect which should be used.
3. When the user taps a Command in the left sidemenu, that sidemenu closes automatically; when the user taps a Command in the right sidemenu, that sidemenu doesn't close automatically: the only way to close is to tap outside the sidemenu.
4. There is an odd graphic defect in the Simulator, as you can see in the video, because the sentence `Contact Us (RIGHT)` is shown as `Contact Us (R  HT)` (without two chars): I cannot try my code in real devices because it cannot compile on your build servers with my new methods (for example, I get: `Exception: java.lang.NoSuchMethodError - com.codename1.ui.Toolbar.addComponentToRightSideMenu`).

The code that I used to test the right sidemenu + the left sidemenu in the same Form (as in the video) is the following:

```
public void start() {
        if(current != null){
            current.show();
            return;
        }
        Form hi = new Form("Hi World", BoxLayout.y());
        hi.add(new Label("Hi World"));
        
        // Test the right sidemenu 
        Toolbar.setOnTopSideMenu(true);
        Toolbar tb = hi.getToolbar();
        Image icon = FontImage.createMaterial(FontImage.MATERIAL_BOOKMARK_BORDER, "SidemenuTagline", 10);
        Button button = new Button(icon, "Label");
        Container topBar = BorderLayout.east(button);
        topBar.add(BorderLayout.SOUTH, FlowLayout.encloseRight(new Label("Help Desk (RIGHT)", "SidemenuTagline")));
        topBar.setUIID("SideCommand");
        tb.addComponentToRightSideMenu(topBar);
        tb.addMaterialCommandToRightSideMenu("Contact Us (RIGHT)", FontImage.MATERIAL_LIVE_HELP, e -> {
            Log.p("Contact Us (RIGHT) tapped");
        });
        
        // Test the left sidemenu (with @hideLeftSideMenuBool=true)
        Image iconLeft = FontImage.createMaterial(FontImage.MATERIAL_BOOKMARK_BORDER, "SidemenuTagline", 10);
        Button buttonLeft = new Button(iconLeft, "Label");
        Container topBarLeft = BorderLayout.west(buttonLeft);
        topBarLeft.add(BorderLayout.SOUTH, new Label("Help Desk (LEFT)", "SidemenuTagline"));
        topBarLeft.setUIID("SideCommand");
        tb.addComponentToSideMenu(topBarLeft);
        tb.addMaterialCommandToSideMenu("Contact Us (LEFT)", FontImage.MATERIAL_LIVE_HELP, e -> {
            Log.p("Contact Us (LEFT) tapped");
        });
        
        
        hi.show();
    }
```

This is my `res/theme.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>

<resource majorVersion="1" minorVersion="9" useXmlUI="false">
    <theme name="Theme">
        <val key="@hideLeftSideMenuBool" value="true" />
        <val key="@includeNativeBool" value="true" />
        <val key="SideCommand.bgColor" value="9966" />
        <border key="SideCommand.border" type="empty" />
        <val key="SideCommand.fgColor" value="ffffff" />
        <font key="SideCommand.font" type="ttf" face="0" style="0" size="0" name="native:MainLight" family="native:MainLight" sizeSettings="3" actualSize="3.5" />
        <val key="SideCommand.margin" value="0.0,1.0,0.0,0.0" />
        <val key="SideCommand.padUnit" value="2,2,2,2" />
        <val key="SideCommand.padding" value="3.0,3.0,3.0,3.0" />
        <val key="SideCommand.press#bgColor" value="6666" />
        <val key="SideCommand.press#derive" value="SideCommand" />
        <val key="SideCommand.sel#bgColor" value="6666" />
        <val key="SideCommand.sel#derive" value="SideCommand" />
        <val key="SideCommand.transparency" value="255" />
        <val key="SideNavigationPanel.bgColor" value="ffffff" />
        <val key="SideNavigationPanel.transparency" value="255" />
        <val key="SidemenuTagline.bgColor" value="9966" />
        <border key="SidemenuTagline.border" type="empty" />
        <val key="SidemenuTagline.fgColor" value="ffffff" />
        <font key="SidemenuTagline.font" type="ttf" face="0" style="0" size="0" name="native:ItalicLight" family="native:ItalicLight" sizeSettings="3" actualSize="3.0" />
    </theme>
</resource>
```

Of course, before merging this code is better to discuss about it :)

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Toolbar.java
+++ b/CodenameOne/src/com/codename1/ui/Toolbar.java
@@ -33,6 +33,7 @@ import com.codename1.ui.events.ActionListener;
 import com.codename1.ui.events.ScrollListener;
 import com.codename1.ui.layouts.BorderLayout;
 import com.codename1.ui.layouts.BoxLayout;
+import com.codename1.ui.layouts.FlowLayout;
 import com.codename1.ui.layouts.Layout;
 import com.codename1.ui.list.DefaultListCellRenderer;
 import com.codename1.ui.plaf.LookAndFeel;
@@ -42,46 +43,54 @@ import java.util.ArrayList;
 import java.util.Vector;
 
 /**
- * <p>Toolbar replaces the default title area with a powerful abstraction that allows functionality ranging
- * from side menus (hamburger) to title animations and any arbitrary component type. Toolbar allows
- * customizing the Form title with different commands on the title area, within the side menu or the overflow menu.</p>
- * 
  * <p>
- * The Toolbar allows placing components in one of 4 positions as illustrated by the sample below:
+ * Toolbar replaces the default title area with a powerful abstraction that
+ * allows functionality ranging from side menus (hamburger) to title animations
+ * and any arbitrary component type. Toolbar allows customizing the Form title
+ * with different commands on the title area, within the side menu or the
+ * overflow menu.</p>
+ *
+ * <p>
+ * The Toolbar allows placing components in one of 4 positions as illustrated by
+ * the sample below:
  * </p>
  * <script src="https://gist.github.com/codenameone/e72cfa6aedd7fcd1af72.js"></script>
  * <img src="https://www.codenameone.com/img/developer-guide/components-toolbar.png" alt="Simple usage of Toolbar" />
- *  
+ *
  * <p>
- * {@code Toolbar} supports a search mode that implicitly replaces the title with a search field/magnifying glass
- * effect. The code below demonstrates searching thru the contacts using this API:
+ * {@code Toolbar} supports a search mode that implicitly replaces the title
+ * with a search field/magnifying glass effect. The code below demonstrates
+ * searching thru the contacts using this API:
  * </p>
  * <script src="https://gist.github.com/codenameone/cd227aaca486889f7c940e2e97985426.js"></script>
  * <img src="https://www.codenameone.com/img/developer-guide/toolbar-search-mode.jpg" alt="Dynamic search mode in the Toolbar" />
- * 
+ *
  * <p>
- * The following code also demonstrates search with a more custom UX where the title
- * area was replaced dynamically. This code predated the builtin search support above. 
- * Notice that the {@code TextField} and its hint are styled to look like the title.
+ * The following code also demonstrates search with a more custom UX where the
+ * title area was replaced dynamically. This code predated the builtin search
+ * support above. Notice that the {@code TextField} and its hint are styled to
+ * look like the title.
  * </p>
  * <script src="https://gist.github.com/codenameone/dce6598a226aaf9a3157.js"></script>
  * <img src="https://www.codenameone.com/img/developer-guide/components-toolbar-search.png" alt="Dynamic TextField search using the Toolbar" />
- * 
+ *
  * <p>
- * This sample code show off title animations that allow a title to change (and potentially shrink) as the user scrolls
- * down the UI.  The 3 frames below show a step by step process in the change.
+ * This sample code show off title animations that allow a title to change (and
+ * potentially shrink) as the user scrolls down the UI. The 3 frames below show
+ * a step by step process in the change.
  * </p>
  * <script src="https://gist.github.com/codenameone/085e3a8fa1c36829d812.js"></script>
  * <img src="https://www.codenameone.com/img/developer-guide/components-toolbar-animation-1.png" alt="Toolbar animation stages" />
  * <img src="https://www.codenameone.com/img/developer-guide/components-toolbar-animation-2.png" alt="Toolbar animation stages" />
  * <img src="https://www.codenameone.com/img/developer-guide/components-toolbar-animation-3.png" alt="Toolbar animation stages" />
- * 
+ *
  * @author Chen
  */
 public class Toolbar extends Container {
 
     /**
      * Indicates whether the toolbar should be properly centered by default
+     *
      * @return the centeredDefault
      */
     public static boolean isCenteredDefault() {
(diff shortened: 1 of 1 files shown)


codenameone:
Thanks! 
Looks great!

Would look better without the format ;-)

>    1. The right sidemenu can be used only as a permanent sidemenu or as an onTop sidemenu, otherwise the methods that I added do nothing.

I don't care about that. on-top mode is the default and the old approach is now deprecated so those are the only two modes that matter.

  > 2.  I didn't implemented the swipe to open and close the right sidemenu: more precisely, in the method constructOnTopRightSideMenu I commented the parent.addPointerDraggedListener(new ActionListener() {...} because the PointerDraggedListener seems to detect only the drag from the left edge to the center, and not from the right edge to the center. I'm also not sure how to manage two PointerDraggedListeners (for the left sidemenu and for the right sidemenu) and how to detect which should be used.

OK, swiping in is indeed a bit of a problem. We can enhance that later.

> 3.    When the user taps a Command in the left sidemenu, that sidemenu closes automatically; when the user taps a Command in the right sidemenu, that sidemenu doesn't close automatically: the only way to close is to tap outside the sidemenu.

I'm guessing it's because of the code in `SideMenuBar` where we wrap side menu commands in a command wrapper. You might need to create a special case for right side menu.
Alternatively I think the better solution is to do this:

- Create a new `closeLeftSideMenu()`
- Have `closeSideMenu()` close both side menus

This makes sense since you wouldn't have both side menus open and having one generic method that closes both makes sense.

> 4.    There is an odd graphic defect in the Simulator, as you can see in the video, because the sentence Contact Us (RIGHT) is shown as Contact Us (R HT) (without two chars): I cannot try my code in real devices because it cannot compile on your build servers with my new methods (for example, I get: Exception: java.lang.NoSuchMethodError - com.codename1.ui.Toolbar.addComponentToRightSideMenu).

This looks like a bug I've seen in right alignment, I think the background of the icon gets painted there due to some reason. These things are a bit hard to track but I don't think they relate to the right side menu.

Some thoughts/enhancements I have about the code in no particular order. We can also merge and do most of these after the merge:

- The code from `addCommandToSideMenu` should move to `addCommandToLeftSideMenu`. We can then implement `addCommandToSideMenu` to delegate to one of these based on the state of RTL

- Can't the code in `addCommandToRightSideMenu(Command cmd)` be generic and common with the left side menu? Most of this code seems like it should be identical with some arguments such as text alignment coming from outside.

- I think the same holds true for `constructOnTopRightSideMenu`

- The pointer drag listener should have a common instance that's used by both side menus, we don't want more than one listener. So it should be refactored to a separate method that's only invoked once

- The same is true for the pointer release/pressed listeners we should have one instance that handles both the left & right side menus

- Theme constants `sideMenuImage` & `sideMenuPressImage` should be `rightSideMenuImage` & `rightSideMenuPressImage`

- Can we make the code of `showOnTopRightSidemenu` & `showOnTopSidemenu` common by passing some arguments to the method?

- In the places where you wrote `// not implemented` we need to throw a runtime exception maybe illegal state exception with a clear message e.g.: "Right side menu bar only works in on-top side menu & permanent mode". Failing fast is better than hiding the problem

- It's possible that `findCommandComponent` will break due to this change so we need to make sure that method takes the right side menu bar into consideration

jsfan3:
Thanks a lot for your precise feedback!
I submitted a new commit that fixes the closing of the right sidemenu in the way you suggested: now the right sidemenu closes automatically when the user taps a Command in the right sidemenu.
I agree to merge now and then do the most of your thoughts/enhancements after the merge.

However, I suppose that the most important enhancement after the merge is to enable the swiping.

codenameone:
The pointer drag listener code is the least important. The most important changes would be ones that cause regressions (which I didn't see) and the ones that introduce method signature changes as those are painful to fix after the fact. The same is true for theme constants etc. those are hard to fix later.
I'll merge this and we'll work from that. Thanks!

PR #2439, opened 2018-06-05
Various enhancements

jsfan3:
I did the best I could to implement the requested enanchments. However my work is not perfect. 
[Three videos in a zip file](https://github.com/codenameone/CodenameOne/files/2070718/videos.zip):
Left and Right sidemenus with swipe
Normal sidemenu without RTL
Normal sidemenu with RTL (it needs fixes)

> The code from addCommandToSideMenu should move to addCommandToLeftSideMenu. We can then implement addCommandToSideMenu to delegate to one of these based on the state of RTL

DONE. I done it also for addMaterialCommandToSideMenu and for addComponentToSideMenu. I kept attention to make all working with a sidemenu that is not in onTop mode or in permament mode (it works correctly) and that is in onTop mode or in permament mode (it needs fixes). I modified addCommandToLeftBar and addCommandToRightBar (the odd fact is that the RTL works with the hamburger icon, but not with the other commands icons). So now the Toolbar has a basic RTL support that should be fixed/improved.

> Can't the code in addCommandToRightSideMenu(Command cmd) be generic and common with the left side menu? Most of this code seems like it should be identical with some arguments such as text alignment coming from outside.

DONE

> I think the same holds true for constructOnTopRightSideMenu

DONE, the true advantage was the possibility to have common instances of the listeners.

> The pointer drag listener should have a common instance that's used by both side menus, we don't want more than one listener. So it should be refactored to a separate method that's only invoked once

DONE, this allowed to implement a working swipe for both the side menus!!! ;-)

> The same is true for the pointer release/pressed listeners we should have one instance that handles both the left & right side menus

DONE

> Theme constants sideMenuImage & sideMenuPressImage should be rightSideMenuImage & rightSideMenuPressImage

DONE

> Can we make the code of showOnTopRightSidemenu & showOnTopSidemenu common by passing some arguments to the method?

Yes, but only the first half of the methods is common. At the moment I leave these methods as they are.

> In the places where you wrote // not implemented we need to throw a runtime exception maybe illegal state exception with a clear message e.g.: "Right side menu bar only works in on-top side menu & permanent mode". Failing fast is better than hiding the problem

DONE

> It's possible that findCommandComponent will break due to this change so we need to make sure that method takes the right side menu bar into consideration

DONE

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Toolbar.java
+++ b/CodenameOne/src/com/codename1/ui/Toolbar.java
@@ -24,6 +24,7 @@ package com.codename1.ui;
 
 import com.codename1.components.InteractionDialog;
 import com.codename1.io.Log;
+import com.codename1.l10n.L10NManager;
 import com.codename1.ui.animations.BubbleTransition;
 import com.codename1.ui.animations.CommonTransitions;
 import com.codename1.ui.animations.Motion;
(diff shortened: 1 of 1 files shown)


codenameone:
Great job!

PR #2443, opened 2018-06-07
Introduce getDeviceDPI and refactor getDeviceDensity()

ramsestom:
This is a smaller PR than my global "deviceResolution refactor" one that only introduce getDeviceDPI(). Other PRs will follow for the DP SP units support or the multiimage DPI resolution support

Remark: I don't know why on earth AndroidImplementation.java is considered as completely changed by git whereas only a few lines changed...


codenameone:
Thanks that's much easier to go through!

I have a few questions/comments:

* We don't use the @Deprecated annotation in Codename One since the port code is a bit problematic and can't use newer features. We only deprecate via JavaDoc

* What happened to `AndroidImplementation` ? It looks like it was reformatted and the diff is very hard to read. I was able to workaround it to see current changes but this might make historic diffs much harder and I'd rather avoid that.

* I still don't understand the value of `getDeviceDPI`?
You're returning `DPI_*` values which means the DPI value is rounded just like the original `getDeviceDensity()` method. So what's the point? This might be more confusing than the original as developers invoke getDeviceDPI() and get an inaccurate value?
I can understand the logic of keeping `getDeviceDensity()` and having a `getDPI()` method that returns a close to accurate value. 
Am I missing something?

Also isn't there a better way to write this:

````java
if ( mID.equalsIgnoreCase("iPad4,4") // iPad Mini 2 Wifi (model A1489)
                      || mID.equalsIgnoreCase("iPad4,5") // iPad Mini 2 Wifi + Cellular (model A1490)
                      || mID.equalsIgnoreCase("iPad4,6") // iPad Mini 2 Wifi + Cellular (model A1491)
                      || mID.equalsIgnoreCase("iPad4,7") // iPad Mini 3 Wifi (model A1599)
                      || mID.equalsIgnoreCase("iPad4,8") // iPad Mini 3 Wifi + Cellular (model A1600)+                      
````

E.g. 

````java
mID = mID.toLowerCase();
if(mID.startsWith("ipad4)) ...
````

ramsestom:
> We don't use the @deprecated annotation in Codename One since the port code is a bit problematic and can't use newer features. We only deprecate via JavaDoc

OK didn't know that. The @Deprecated annotation isn't really a new feature of java though. It is here since J2SE 5.0... 

>  What happened to AndroidImplementation ? It looks like it was reformatted and the diff is very hard to read. I was able to workaround it to see current changes but this might make historic diffs much harder and I'd rather avoid that.

I really don't know. This new PR has been a real nightmare to create as git was completely lost by the fact you merged and then reverted my first PR. So I had to delete all the files affected by this PR, refetch them from your master branch and reedit them one by one by hand to put my changes. The process for `AndroidImplementation` was the same than for the others so I really don't know why github consider it as complettely changed (when I force him to load the diff in the "Files changed" section of this PR I can't see any difference for the line 1 of the two files for example whereas it consider them as a deletion and an addition...). 

> I still don't understand the value of getDeviceDPI?
> You're returning DPI_* values which means the DPI value is rounded just like the original getDeviceDensity() method. So what's the point? This might be more confusing than the original as developers invoke getDeviceDPI() and get an inaccurate value?
> I can understand the logic of keeping getDeviceDensity() and having a getDPI() method that returns a close to accurate value.
> Am I missing something?

Yes you are missing the fact that DPI_* values are returned only in the default CodeoameOneImplementation . For platforms where whe have a more accurate information (like android and ios), the getDeviceDPI method is overrided and this is this more accurate value that is returned (so it would be an exact value on iOS and a close to reality value on android)
Of course that for platforms where we have no other information than the screen resolution we are forced to return an estimate of the screen density.

> Also isn't there a better way to write this:
> 
> if ( mID.equalsIgnoreCase("iPad4,4") // iPad Mini 2 Wifi (model A1489)
>                       || mID.equalsIgnoreCase("iPad4,5") // iPad Mini 2 Wifi + Cellular (model A1490)
>                       || mID.equalsIgnoreCase("iPad4,6") // iPad Mini 2 Wifi + Cellular (model A1491)
>                       || mID.equalsIgnoreCase("iPad4,7") // iPad Mini 3 Wifi (model A1599)
>                       || mID.equalsIgnoreCase("iPad4,8") // iPad Mini 3 Wifi + Cellular (model A1600)+                      
> E.g.
> 
> mID = mID.toLowerCase();
> if(mID.startsWith("ipad4)) ...

Yes assuming that mID would never be returned in upper case (I don't know if this is possible). 
Anyway device DPI value is cached so this test would only be performed once (so there won't really be any performance penality of doing the test this way rather than using a more concise one ) and I thought it was clearer that way (when new ios device appear, it is easier to see where you should add them. For example if there is an iPad Mini 6 with the same resolution as iPad Mini 3 but with a mID that do not start with iPad4, this is easier to add its individual mID test rather than having to rethink the test logic...) .

codenameone:
So here's what I think:

- I want to undeprecate getDeviceDensity() it works as it always did and I don't see a reason to change that. 

- I don't think the DPI* constants make sense. They would just be confusing, we shouldn't have them.

Notice that in my code I started it with `mID = mID.toLowerCase();` so it would always be lower case.

codenameone:
The PPI method returns an approximation that is slightly more accurate but you only looked at two ports and ignored device bugs.

The `getDeviceDensity` makes more sense for a lot of use cases since it returns a predictable set of values that you can switch/case. Having a PPI value would fit in this scenario only if it returned "fake" rounded values and that's confusing (hence my problem with the DPI constants). Android does this exact thing where it returns a value matching a set of constants. The fact that these constants map to a PPI value is mostly for convenience of the Android API developers.

Our exotic unit of measurement is well tested and documented, it isn't different from the way Android or Apple abstract the PPI values. Hell Apple doesn't publish the PPI values at all so the idea of this being familiar isn't accurate. Furthermore, this isn't a feature that's generally exposed. Most developers just use the high level API's. 
I accept that having DIPs, inches etc. as a unit of measurement makes sense but those are auxiliary options and unrelated to this change. Having a DPI method makes sense too but that should (and does) return an accurate value hence having constants for it or replacing the usage of `getDeviceDensity` doesn't fit. These are different concepts for different purposes.

We have something that's working, documented & tested on 7 ports... It's familiar to Android developers and relatively low level so familiarity isn't crucial at that level. I accept the value of the other changes.

Yes I missed that nuance in the if statement. I'm guessing we can probably do better there. I also missed the ipad 5 constants. Maybe the logic should map to an array.

ramsestom:
Closing this as it as been innactive for 1 month

PR #2450, opened 2018-06-13
added a button constructor that takes a command and a uiid as arguments and improved uncover transition

Firethunder:


codenameone:
Can you explain why you made that change to the cover transition?

This is uncover so it's supposed to go in he opposite direction.

Firethunder:
Hi,

looks like I ran into some kind of an edge case with the uncover transition. 
If it is used in a form transition the uncover animation works well, but if it is used with Container.replaceAndWait it does not show the uncovered component until the animation is finished.
Moving up the paintBackgrounds call seems to fix this.
[example code](https://gist.github.com/Firethunder/b43dd4a582ee7fbb3be0e08449ec05eb)

codenameone:
Doesn't this break the behavior of forms though?

shai-almog:
@all-contributors please add @Firethunder  for code

Issue #2458, opened 2018-06-20
RFE: show the cause of a validation error on TextModeLayout (on iOS)


jsfan3:
_Related question: https://stackoverflow.com/questions/50886775/codename-one-textmodelayout-on-ios-and-validation-error-feedback_

_I copy the question for reference:_

The `TextModeLayout`, on iOS, produces only a red cross on the right of a text field to indicate an invalid input (if we use a `Validator`).

In general, this is fine. However, I added a `LengthConstraint(8)` to a password field and I'm worried if users understand that there is a requirement for lengtness.

Is there a standard way in the iOS design to provide the user the cause of a validation error, or the only way is to use the Material Design also on iOS (that means set to `true` the theme const `textComponentOnTopBool`)?

Code:

       TextModeLayout tl = new TextModeLayout(2, 1);
        Container credentials = new Container(tl);
        credentials.setUIID("MyContainer01");
        TextComponent email = new TextComponent().label("FieldEmail");
        TextComponent password = new TextComponent().label("FieldPassword");
        Validator val = new Validator();
        val.addConstraint(email, RegexConstraint.validEmail());
        val.addConstraint(password, new LengthConstraint(8));
        val.addSubmitButtons(loginButton);
        email.getField().setConstraint(TextField.EMAILADDR);
        password.getField().setConstraint(TextField.PASSWORD);
        InputComponent.group(email, password);
        credentials.add(email);
        credentials.add(password);

codenameone:
Do you have a suggestion for how this should look?

jsfan3:
Yes, now I have a suggestion for how it should look. 

I implemented this RFE with a preliminary code, as you can see in this video:
https://jmp.sh/sFYEp5g

Basically, the error message is shown for two second in place of the label on the left of the InputComponent (or on right of the InputComponent for RTL languages). This solution never breaks the layout, because the error message is trimmed to fit the available space.

If you would like to test (and to comment and/or to improve) my code, replace in your code `Validator val = new Validator();` with `Validator val = new ExtendedValidator();` and, of course, import my `ExtendedValidator` class. That's all.

Code:

```
import com.codename1.ui.Component;
import com.codename1.ui.Display;
import com.codename1.ui.InputComponent;
import com.codename1.ui.Label;
import com.codename1.ui.PickerComponent;
import com.codename1.ui.TextComponent;
import com.codename1.ui.events.FocusListener;
import com.codename1.ui.util.UITimer;
[...]

PR #2475, opened 2018-07-12
Show errors when is not onTopMode

jsfan3:
Implement this RFE: https://github.com/codenameone/CodenameOne/issues/2458


codenameone:
Thanks!

codenameone:
FYI I missed the exception that limited addConstraint to input components. This was a mistake.
I fixed it in git.

jsfan3:
I'm sorry for that. Thank you for the fix and for the news in the Codename One blog.

PR #2548, opened 2018-09-17
added getter for url

kutoman:
just for the case of debugging. When e.g. a request fails and the url of the request should be logged


codenameone:
Thanks.
BTW please try to include a JavaDoc comment on the added methods otherwise the generated JavaDoc looks "sad" ;-)

kutoman:
ok I'll keep that in mind for the future :)

PR #2576, opened 2018-10-11
Fix bug of stroke thickness not taken into account when stroke passed…

ramsestom:
… as an object in RoundRectBorder


codenameone:
Thanks!

PR #2592, opened 2018-10-23
Added method to zoom to a specific location panning the image. Also, zoom is now animated

carlosverdier:


codenameone:
There are two fixes I'd like to change here. 

- The current zoom method doesn't animate and you don't have a toggle option here. This is a problem. We need a flag to enable animate on zoom, it can be on by default but there should be a way to turn it off

- I don't understand why we need the `AnimateZoom` class? 
The `ImageViewer` is already animated, why not incorporate that functionality into the enclosing class?

carlosverdier:
I saw the class AnimatePanX and thought it was the way to go...
Hope it's fine now.

shai-almog:
@all-contributors please add @carlosverdier  for code

PR #2614, opened 2018-11-22
escape from getSomeData if there will never be data

ddyer0:
If the remote end closed the inputstream while the local thread is
in getSomeData, the local end hasn't been closed, and the thread
hangs forever.

### DIFF
--- a/CodenameOne/src/com/codename1/io/Socket.java
+++ b/CodenameOne/src/com/codename1/io/Socket.java
@@ -221,7 +221,8 @@ public class Socket {
           		// end has closed the socket, resulting in eof, but the local end hasn't
           		// closed it yet.
           		if(buffer==null) { return(false); }
-        		if( !closed
+        		if( (buffer.length==0)
+        			&& !closed
         			&& Util.getImplementation().isSocketConnected(impl))
          		{	// wait a while if there's still hope
         			try {


codenameone:
What about the case of `buffer.length == 0` ?

ddyer0:
Buffer length=0 is a legal state, if a previous read exactly emptied the buffer and no more
data has arrived yet.

codenameone:
Length 0 is valid but after this change it will no longer sleep in the loop

ddyer0:
you're right, add buffer.length==0 && to the if clause

ddyer0:
Networking on IOS seems to be completely broken.  Unless you changed something else,
this is the immediate suspect, though I probably shouldn't be fixed by simply reverting this.

ddyer0:
I think the proper fix is to make the IOS implementation conform the the same contract
as android, so in IOSImplementation.java

`	static final byte[]dummyval = new byte[0];
   @Override
    public byte[] readFromSocketStream(Object socket) {
        byte []val = nativeInstance.readFromSocketStream(((Long)socket).longValue());
        if(val==null) { return(dummyval); }
        return(val);
    }
    `
The same problem may exist for javascript

codenameone:
I reverted the PR for now

PR #2742, opened 2019-03-18
Update RoundRectBorder.java

kutoman:
fixed: bottomOnlyMode, topOnlyMode causing the opposite effect


codenameone:
This is problematic as it will break compatibility. This functionality is deprecated anyway as we have granular control over corners

PR #2755, opened 2019-04-05
add a wasNull() function to the database Row interface

ramsestom:
add a wasNull() function to the database Row interface allowing to test whether the last column read had a value of SQL NULL

### DIFF
--- a/CodenameOne/src/com/codename1/db/Row.java
+++ b/CodenameOne/src/com/codename1/db/Row.java
@@ -101,4 +101,15 @@ public interface Row {
      * @throws IOException 
      */
     public String getString(int index)throws IOException;
+    
+    
+    
+    /**
+     * Reports whether the last column read had a value of SQL NULL. 
+     * Note that you must first call one of the getter methods on a column to try to read its value and then call the method wasNull to see if the value read was SQL NULL.
+     * 
+     * @return true if the last column value read was SQL NULL and false otherwise. If the targeted platform do not support this feature it returns null (meaning we don't now if the latest read value was SQL NULL or not)
+     * @throws IOException 
+     */
+    public Boolean wasNull()throws IOException;
 }
--- a/CodenameOne/src/com/codename1/db/ThreadSafeDatabase.java
+++ b/CodenameOne/src/com/codename1/db/ThreadSafeDatabase.java
@@ -218,6 +218,14 @@ public class ThreadSafeDatabase extends Database {
             });
         }
         
+        
+        public Boolean wasNull()throws IOException{
+             return (Boolean)invokeWithException(new RunnableWithResponseOrIOException() {
+                public Object run() throws IOException {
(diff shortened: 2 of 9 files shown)


codenameone:
This seems to be missing the UWP version and we'll need to add a JS version of this.

ramsestom:
Added the UWP version. Unsure of how C# nullable bool (bool?) would translate into a java Boolean object though. You will have to check this.

codenameone:
@shannah can you review?

shannah:
Why `Boolean wasNull()` and not `boolean wasNull()`?

ramsestom:
Because it allows to differentiate the cases where this function is not implemented for the targeted platform (like for blackberry) from cases where the SQL value was actually NULL or not.
So `if ( true == wasNull() )` it means the SQL value was actually NULL, `if (false == wasNull() )`, the SQL value was not null and `if (wasNull() == null)` we don't know because the platform do not support this functionnality or it wasn't implemented.
The other solution would be to return a `boolean ` and have an additional function like `supportWasNull() ` that also return a boolean and gives the information wether `wasNull()` is supported by the given platform or not. But I think it is easier to have all the information encoded into the result of a single function (hence the `Boolean` object that allows to encode 3 states (true, false, null) rather than a `boolean` ) and just have to call one function for each test rather than 2 consecutive functions each time we want to know if an SQL value was NULL or not...

shannah:
But won't `if ( true == wasNull() )` throw an NPE if `wasNull()` returns null?  That would mean a lot of protective coding around these parts making the API more difficult to use.

ramsestom:
Yes you are right. I thought that the JVM would automatically cast a boolean to a Boolean object when compared to another object (null) but it isn't the case. Actually you need to write the test like this `if (Boolean.TRUE.equals(wasNull()))` so that no NPE is thrown

So If I writte:
```
Boolean b = null;
if (Boolean.TRUE.equals(b)) {
	System.out.println("TRUE");
}
if (Boolean.FALSE.equals(b)) {
	System.out.println("FALSE");
}
if (b == null) {
	System.out.println("NULL");
}
```
it works as expected and prints "NULL"


if you prefer the solution with two functions (wasNull() and supportsWasNull() or implementsWasNull()) returning `boolean`s you can just modify this PR though. I don't mind.

shannah:
I would have the default implementation throw a RuntimeException with a message clearly stating why it is thrown.  And return `boolean`.  This kind of method needs to have 100% coverage anyways so the default implementation should never be called.

ramsestom:
Thing is there is no default implementation as Row is an interface not a class nor an abstract class. So each platform has its own implementation but there is no default one. else I would have take this approach. But if you prefer to transform the row interface into an abstract class you can.

codenameone:
@shannah should this be merged?
I think that since this method isn't crucial it doesn't matter if it isn't universally supported as long as there's a stub implementation and it's well documented.

Issue #2764, opened 2019-04-13
[BUG] style modifications lost when hidding and unhidding a component


ramsestom:
using `component.setHidden(true);` and then `component.setHidden(false);` makes you loose all the modifications that have been applied to the styles of a component. 
This is because the component is unhidden with  `setUIID(getUIID()); ` that restore all the defaults style values from the style assotiated with this UIID (and not just the margin that might also have been changed from the defaults anyway). 
Rather than dooing this, it would make more sense to cache the margins before the component was hidden into the component object (or into the style object) when `setHidden(true)` is called and to restore them to theses values if `setHidden(false)` is called.

PR #2765, opened 2019-04-14
Fix bug of style modifications beeing lost when hidding/unhidding a comp

ramsestom:
Fix this: https://github.com/codenameone/CodenameOne/issues/2764

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Component.java
+++ b/CodenameOne/src/com/codename1/ui/Component.java
@@ -6778,15 +6778,15 @@ public class Component implements Animation, StyleListener, Editable {
         if(b) {
             if(!sizeRequestedByUser) {
                 if(changeMargin) {
-                	getAllStyles().setNullMargins();
-                    //getAllStyles().setMargin(0, 0, 0, 0);
+                	getAllStyles().cacheMargins(false); //if a margins cache already exists because the component is already hidden it would be kept else it would be created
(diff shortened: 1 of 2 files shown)


codenameone:
Why did you add two public methods that are deprecated? 
Wouldn't it make more sense to create a proper API that makes sense for the general use case instead of internal public API?
This can also pose a problem if `setHidden(true)` is invoked twice. That can happen in real life if set hidden is mapped to a UI element and the user presses it twice. It would result in hard to reproduce issues.

ramsestom:
removed the deprecated tag but the user must be warned that unexpected results might happen when hidding/unhidding a component if he mess with this margins cache... (as long as this cache can be accessed by public methods, there is no other solution)
also calling `setHidden(true)` multiple consecutive times shouldn't be an issue anymore now

codenameone:
I don't understand how calling `setHidden(true)` multiple times is solved here. What did I miss?

ramsestom:
the `cacheMargins` now takes a boolean argument that, when set to false, will only erase the margins cache if this cache do not already exists. When hidding an element, this function is called with a `false` parametter. So if you hidde an element multiple times, the margin values would be cached only the first time (when they are not equals to zero exept if your element had null margins before beeing hidden of course) and restored to these values when unhidden.

codenameone:
Thanks!

PR #2795, opened 2019-05-11
Autosize minimum font

ThomasH99:
Currently, there is no minimum for AutoSize so with long texts the font may become so small it is unreadable. This update mirrors the current support for Maximum font size.


codenameone:
Thanks

PR #2855, opened 2019-06-26
Fixed scrolling down when dragging hidden comp

ThomasH99:
Hi, I think I found a solution which should not have any side-effects for the issue that you cannot drag a hidden component down (discussed here in SO: https://stackoverflow.com/questions/55816378/how-to-correctly-implement-fix-to-make-dragging-a-hidden-component-downwards-wor?noredirect=1#comment98948835_55816378). Hope it's acceptable, otherwise I hope you can help me find a better fix.


codenameone:
Can you resolve the conflicts in the PR?

ThomasH99:
I'm unfortunately no master of Git/Github, but when I look now, it says "This branch has no conflicts with the base branch" and I can't find a way to display any conflicts. Has it somehow resolved itself?

codenameone:
Thanks!

PR #2856, opened 2019-06-26
Add set font icon

ThomasH99:
Hi, I made an attempt to add support for Font Icons (in addition to built-in support for Material Icons). Primarily by generalizing the support you already had for Material icons. 

As discussed recently in the Blog: https://www.codenameone.com/blog/using-icon-fonts-such-as-fontello.html


codenameone:
You can't remove the setMaterialIcon API's. These are used all over the place. You can't just rename them to add your functionality. This breaks compatibility which is something we don't do.

You can add another method and/or enhance functionality. E.g. setMaterialIcon can use a font icon if it's an icon font is explicitly defined for the component.

Also notice that the set methods in the `ImageFont` class predated the API for material icon in label. Had that API existed back then they would have used that API. So when you create a new API you should probably use the API in label and add an API to `ImageFont` just as a delegate to that API.

ThomasH99:
I think you didn't review the changes in detail. I did not remove or change any of the existing APIs, I only reshuffled some code from the existing material icon methods to the set font icon methods to avoid duplication. Let me know if I missed something or misunderstood your answer.

codenameone:
Thanks!

You're right. I got stumped by a merge highlight that looked wrong and missed the code that fixed it below.

ThomasH99:
Happy to be able to contribute :-)

PR #2868, opened 2019-07-24
240719 0505 pm possibility to listen out of bounds in Interaction dialog

DurankGts:
Added Possibility to listen out of bounds in Interaction Dialog to take some action


codenameone:
This removes all recent commits we made as well as conflicts with existing code.
Also the API doesn't follow our convention of adding/removing a listener using `EventDispatcher`.

shai-almog:
@all-contributors please add @DurankGts  for code

PR #2870, opened 2019-07-25
250719_1021_am possibility to listen out of bounds in interaction dialog

DurankGts:
Added possibility to listen out of bounds in interaction dialog


codenameone:
This still uses the `Runnable` approach instead of the `ActionListener` / `EventDispatcher` approach for binding/unbinding a callback.

DurankGts:
please check the new update

codenameone:
See `EventDispatcher`, add/remove listener pattern.

PR #2901, opened 2019-08-31
Returs the components in the correct order

jsfan3:
The commit https://github.com/codenameone/CodenameOne/commit/2ef6a77e4f8ac3475a8c05541b8ac829099ac9a9 is incomplete and doesn't fix the issue https://stackoverflow.com/questions/57734646/componentselector-doesnt-return-components-in-the-correct-order/57735350. This commit completes the previous commit, fixing the issue: "ComponentSelector doesn't return Components in the correct order".


jsfan3:
Why have all check failed?

codenameone:
Thanks!
We're reworking our CI infrastructure, hopefully we'll have something better soon.

PR #2903, opened 2019-09-03
Javadoc fix

jsfan3:
Fix of the Javadoc according to our discussion on Stack Overflow:
https://stackoverflow.com/questions/57744724/codename-one-stringbuilder-vs-operator-between-strings/57750844#57750844


codenameone:
Thanks!

PR #2904, opened 2019-09-03
TextField.NON_PREDICTIVE

jsfan3:
Because it's a password, it should be better TextField.NON_PREDICTIVE than TextField.ANY


codenameone:
Thanks!

PR #2910, opened 2019-09-06
add a generic Interface for releasable Components

ramsestom:
Had to define my own custom buttons and other interactive components so I made a generic Interface for releasable components that any Component that needs to be released by the parent Form can implement. 
You are free to add this modification to CN1 if you think it can be usefull to others.


codenameone:
Thanks!
I'll change the `IRelease` to match our conventions better

shai-almog:
@all-contributors please add @ramsestom  for code

PR #2915, opened 2019-09-11
Some fixes

jsfan3:
Motivation: http://docs.oracle.com/javase/6/docs/api/java/text/DateFormat.html doesn't provide examples of patterns, while https://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html provides examples and explanations that (mostly) apply to this class. I also added a link to the discussion on Stack Overflow.
The Javadoc of the parse() method was wrong because it wasn't a Javadoc (note that it started with /* instead of /**) and so it resulted wrongly as "not implemented" in your official Javadoc (because the Javadoc of the overridden parse() method of the DateFormat class reports that it's not implemented). I fixed it with an actual Javadoc. I also added an exception that is more significant than the exception thrown by the superclass if a template is not provided.

### DIFF
--- a/CodenameOne/src/com/codename1/l10n/SimpleDateFormat.java
+++ b/CodenameOne/src/com/codename1/l10n/SimpleDateFormat.java
@@ -32,9 +32,11 @@ import java.util.Vector;
 
 /**
  * A class for parsing and formatting dates with a given pattern, compatible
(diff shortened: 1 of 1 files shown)


codenameone:
Thanks!

PR #2925, opened 2019-09-18
Added methods to guess mime types

jsfan3:
Codename One lacks of methods to guess mime types (OpenJDK8 provides them). Sometimes they could be useful: for example, my Android phone returns files from Gallery without extension. Checking if a given file is an Image, for example, is a lot faster and without RAM usage trying to guess the mime type than trying to instantiate an Image object.

Currently supported mime types:

application/java-vm, application/msword, application/pdf, image/photoshop, application/postscript, audio/mp3, application/x-java-serialized-object, text/html, application/xml, image/bmp, image/tiff, image/gif, image/x-bitmap, image/x-pixmap, image/png, image/jpeg, image/jpg, application/acad, audio/basic, audio/x-wav, application/zip

In future we can add other mime types with other Pull Requestes.
Note that my implementation is similar but not equal to the OpenJDK8 implementation: inside the code, I inserted two links for reference.

### DIFF
--- a/CodenameOne/src/com/codename1/io/Util.java
+++ b/CodenameOne/src/com/codename1/io/Util.java
@@ -1827,7 +1827,7 @@ public class Util {
             inputStream = Storage.getInstance().createInputStream(sourceFile);
         }
 
-        return MimeTypes.guessMimeType(inputStream);
+        return guessMimeType(inputStream);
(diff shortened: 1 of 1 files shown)


codenameone:
Thanks!

PR #2986, opened 2020-01-07
added anchors for percent based insets

kutoman:
With these anchors percent based insets can also refer to other points than the default one (e.g. center of a Component) .


shai-almog:
@all-contributors please add @kutoman  for code

PR #2998, opened 2020-01-18
RFE: Display.invokeWithoutBlockingWithResultSync

jsfan3:
This morning I was reading again the blog post https://www.codenameone.com/blog/invoke-without-blocking.html to add `CN.invokeWithoutBlocking` to my code. It's very interesting, but I'm not sure if it's applicable to the critical code of my app, because that code has to return an object. That's why I'm adding the method `invokeWithoutBlockingWithResultSync`.


codenameone:
Thanks!

PR #3015, opened 2020-02-01
Added "see also" link in the javadoc

jsfan3:
I added a link to a blog article of Steve that clarify the externalizable interface in a way deeper than the Javadoc.


codenameone:
Thanks!

Issue #3025, opened 2020-02-13
[BUG + FIX] fetchAsProperties throws an exception if the PBO contains a ListProperty


jsfan3:
Assuming to have a PBO like this:
```
public class ChatGroupDAO implements PropertyBusinessObject {
    
    public final Property<String, ChatGroupDAO> id = new Property<>("id", String.class);
    public final Property<String, ChatGroupDAO> groupName = new Property<>("groupName", String.class);
    public final Property<String, ChatGroupDAO> adminUniqueID = new Property<>("adminUniqueID", String.class);
    public final ListProperty<String, ChatGroupDAO> membersUniqueIDs = new ListProperty<>("membersUniqueIDs", String.class);
    
    private final PropertyIndex idx = new PropertyIndex(this, "ChatGroupDAO",
            id, groupName, adminUniqueID, membersUniqueIDs
    );
    
    @Override
    public PropertyIndex getPropertyIndex() {
        return idx;
    }
    
}
```

and to use in the app code the following method of RequestBuilder:
```
public ConnectionRequest fetchAsProperties(final OnComplete<Response<PropertyBusinessObject>> callback, final Class type)
```

if the server returns the following JSON, that maps perfectly on ChatGroupDAO, we have a casting exception (impossible to cast a String to a PropertyBusinessObject):

```
{
   "id":"MDliSqPjLwiqKYnRAkG0YPHVzlkQ4c9izayimeangDkn56zZQEoDAbqnC7O2hjXN",
   "groupName":"Mitici",
   "adminUniqueID":"QyVNImtVbJhceXr5Ec4Ef7sjFYziDALK6grjcPEw2q0PpkBARfZ8MeKGHLJNbGKH",
   "membersUniqueIDs":[
      "eqGaFwKr77KbSAlkg41DugLUXMSYdkJsCCdwdPZEtx0QbOZVulDhNEPUTCiSnRHw",
      "PfsOgxTaN7MJuGPUd38PN7jsABrqBN22ihnAUsUamUoBPZfsZIfGnFhwfIm8wO81"
   
    ]
}
```

It's not easy for me to provide you a full working test case because this issue is not reproducible without posting both client code and server code, however I've already found the cause of the bug and a possible fix.

In the file `PropertyIndex.java`, the problematic lines are from 342 to 356:
```
List l = (List)val;
if(!l.isEmpty()) {
    if(l.get(0) instanceof PropertyBusinessObject || p.getGenericType() == null) {
[...]

PR #3026, opened 2020-02-13
Fix https://github.com/codenameone/CodenameOne/issues/3025, fetchAsPr…

jsfan3:
…operties throws an exception if the PBO contains a ListProperty

Fix https://github.com/codenameone/CodenameOne/issues/3025, fetchAsProperties throws an exception if the PBO contains a ListProperty


codenameone:
Thanks!

PR #3028, opened 2020-02-15
SpanMultiButton implementation

jsfan3:
`SpanMultiButton` was missing: I needed it, so I created it. I used your `MultiButton` implementation as base.
The javadoc contains a code example as test case and two images of the Simulator.

The gist that I used in the Javadoc and the images are here: https://gist.github.com/jsfan3/83586d7b6db9b66772d02d25898f1bad

### DIFF
--- a/CodenameOne/src/com/codename1/components/SpanMultiButton.java
+++ b/CodenameOne/src/com/codename1/components/SpanMultiButton.java
@@ -30,6 +30,7 @@ import com.codename1.ui.Container;
 import com.codename1.ui.Image;
 import com.codename1.ui.Label;
 import com.codename1.ui.RadioButton;
+import com.codename1.ui.TextArea;
 import com.codename1.ui.events.ActionListener;
 import com.codename1.ui.events.ActionSource;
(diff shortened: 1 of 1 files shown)


codenameone:
Thanks!
Won't it make more sense to use `TextArea` instead of `SpanLabel` to prevent depth and UIID hell with the additional redundant hierarchy?

Did you try running it with very long text, it might look pretty weird. Won't it make sense to block the places where it can span and the levels of which it can/should span?

jsfan3:
Yes, in this case the use of `TextArea` is better than `SpanLabel`, my last commit implements this change.

About the use of a very long text, I tried this test case:
https://gist.github.com/jsfan3/4233b19601a69cde650330808e68df3c
however, the only issue that I noted is a not perfectly fluid scroll on iOS real device and on Simulator with iOS skin, while on a real Android device and on Simulator with Android skin is fine.
I suspect that this issue is not related to the code of `SpanMultiButton`, because I have a similar problem in other code.

What do you mean for "to block the places where it can span and the levels of which it can/should span"?

codenameone:
Thanks!

PR #3073, opened 2020-04-05
Better documentation and a new code example

jsfan3:
Currently the javadocs and the developer guide indicate only the `Capture` API to record audio. A complete example without using the `Capture` API is missing and the `Media` javadoc is misleading in some points. This commit fixes the javadoc, adding a complete customizable example of audio recording without the need of the `Capture` API.


codenameone:
Thanks!

PR #3075, opened 2020-04-09
Fix a NPE

jsfan3:
The use of `cmp.getComponentForm()` assumes that the the component to be validated is already inserted in a Form. This is not necessary true. For example, it may happen that some InputComponents (binded to a Validator) are inserted in a Form with already some data: in this case, the Validator is called before the Form is shown, throwing a NullPointerException. This happens in a particular circumstance in the app I'm developing. This commit fixes this issue.


codenameone:
Thanks!

PR #3093, opened 2020-05-05
Swipeable autoclose

ThomasH99:
Added support necessary to close an already open SwipeableContainer as soon as another one starts opening.


CodenameOne/src/com/codename1/ui/SwipeableContainer.java
```diff
@@ -383,6 +383,8 @@ public void actionPerformed(ActionEvent evt) {
                     }
 
                     if (initialX != -1) {
+                        if (getPreviouslyOpened() != null && getPreviouslyOpened() != SwipeableContainer.this && getPreviouslyOpened().isOpen())
+                            getPreviouslyOpened().close();
```
codenameone, line 387: Can you add curly brackets around the statement?

CodenameOne/src/com/codename1/ui/SwipeableContainer.java
```diff
@@ -444,4 +446,12 @@ public void actionPerformed(ActionEvent evt) {
...
     }
 
+    /**
+     * override to return a previously opened SwipeableContainer that should be automatically closed when starting to open this one
+     * @return 
+     */
+    public SwipeableContainer getPreviouslyOpened() {
+        return null;
+    }
+
```
codenameone, line 456: Can you add a field and a setter, also include text in the `@return` tag otherwise the Javadoc looks weird

ThomasH99: In my use case (and I believe most other), a field and setter doesn't make sense: I have a long list of SwipeableContainers and keep track of which is opened in the Form (I overwrite openToLeft, openToRight and close() to store/remove an open Swipeable). For a local field (like 'openSwipeable') to make sense, every SwipeableContainer would need to get that set. 

I assume there might be a better overall way to smoothly close any already opened Swpeiable, but here I went for minimal impact to the existing code. 

(NB. My first attempt was to override openToLeft/openToRight to close the previously opened Swipeable, but since they only get called once Swipeable is completely opened, it didn't look right).

If you still think it makes sense, I can add these since it won't break anything, at worst it could lead a developer down a wrong path at first :-)

CodenameOne/src/com/codename1/ui/SwipeableContainer.java
ThomasH99: Thanks, yes, sure

PR #3096, opened 2020-05-06
Definitive solution for image orientation issues

jsfan3:
Example code that solves the image orientation issues reported on https://stackoverflow.com/q/61619714:
https://gist.github.com/jsfan3/7fc101523955e8179fadd2c713a09e05

Tested on my devices. See the javadocs :)

The hard work consisted of main issues: the first was find a way to detect the Exif Orientation Tag without any external library; the second was understand how to manipulate the image according to the orientation tag; the third was make the code efficient. For the latter issue, the only valid solution that I found is to add a maxSize value, as explained in the javadoc and demostrated by the code example.

I hope this could be a definitive solution. Moreover, it could be interesting for a lot of developers, so this new method could be reported in the Codename One blog.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Image.java
+++ b/CodenameOne/src/com/codename1/ui/Image.java
@@ -25,6 +25,7 @@ package com.codename1.ui;
 
 import com.codename1.ui.geom.Dimension;
 import com.codename1.impl.CodenameOneImplementation;
+import com.codename1.io.FileSystemStorage;
 import com.codename1.io.Log;
 import com.codename1.io.Util;
 import com.codename1.ui.events.ActionEvent;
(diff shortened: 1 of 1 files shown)


CodenameOne/src/com/codename1/ui/Image.java
```diff
@@ -1207,4 +1208,347 @@ public void fireChangedEvent() {
...
+    /**
+     * Very fast method to detect if the given inputStream is a PNG image
+     * (according to its guessed mime type)
+     *
+     * @param inputStream
+     * @return true if PNG, false otherwise
+     */
+    public static boolean isPNG(InputStream inputStream) throws IOException {
+        String type = Util.guessMimeType(inputStream);
+        if ("image/png".equals(type)) {
```
codenameone, line 1548: shouldn't this be `return "image/png".equals(type);`?

CodenameOne/src/com/codename1/ui/Image.java
```diff
@@ -1207,4 +1208,347 @@ public void fireChangedEvent() {
...
+     * @param rotatedImage is the FileSystemStorage path in which the capture
+     * photo is stored, normally this should be inside the
+     * FileSystemStorage.getAppHomePath().
+     * @param maxSize is the maximum value of the width and height of the
+     * rotated images, that is scaled if necessary, keeping the ratio.
+     * @return the com.codename1.ui.Image
+     * @throws java.io.IOException
+     */
+    public static Image exifRotation(String capturedImage, String rotatedImage, int maxSize) throws IOException {
+        FileSystemStorage fss = FileSystemStorage.getInstance();
```
codenameone, line 1286: This API currently only works with file system storage. It would be great if it can work with a header byte array.

CodenameOne/src/com/codename1/ui/Image.java
```diff
@@ -1207,4 +1208,347 @@ public void fireChangedEvent() {
...
+        boolean isJpeg = isJPEG(fss.openInputStream(capturedImage));
+        boolean isPNG = isPNG(fss.openInputStream(capturedImage));
```
codenameone, line 1288: Here to input streams are opened without closing re-reading the same data. If this worked with the byte array data this could be reused.

CodenameOne/src/com/codename1/ui/Image.java
```diff
@@ -1207,4 +1208,347 @@ public void fireChangedEvent() {
...
+        Image img = EncodedImage.create(fss.openInputStream(capturedImage), (int) fss.getLength(capturedImage));
+        if (maxSize > 0 && (img.getWidth() > maxSize || img.getHeight() > maxSize)) {
+            // Tested that scaling the image before rotating is a lot more efficient than rotating before scaling
+            img = img.scaledSmallerRatio(maxSize, maxSize);
+        }
+        switch (orientation) {
+            case 0:
+            case 1:
+                // no rotation
+                ImageIO.getImageIO().save(img, fss.openOutputStream(rotatedImage), format, 0.9f);
```
codenameone, line 1312: Why write the image if there's no orientation change?

jsfan3:
Dear Shai, feel free to modify my code as you think it’s better. Your understanding of Java and Codename One is far superior to mine.  Just one note: using ImageIo.save even when the image is not rotated is because it might have been resized if maxSize was specified: this way the developer knows that the image is always resized if necessary.  Having said that, please fix the code as you think it’s better.  A hug and thanks!

codenameone:
Makes sense. Thanks!

PR #3100, opened 2020-05-13
Swipeable autoclose updated

ThomasH99:
Updated based on Shai's comments


ThomasH99:
Updated based on Shai's comment on previous PR

codenameone:
Thanks!

PR #3123, opened 2020-05-28
Fixed error in boolean returned value

jsfan3:
The value returned by `delete()` must be `true` if the file does not exist, but in the current implementation it is the opposite. This fix allows the current implementation to be corrected.


codenameone:
Thanks!

PR #3130, opened 2020-06-02
Added test for getComponentForm() returning null pointer

ThomasH99:
Added test of whether getComponentForm return null (this test is also done in other components). This caused a crash when moving between forms with animation


codenameone:
You left the original code commented out. But here's the thing, getComponentForm should never ever return null in init. The only way this can happen is with an EDT violation.

ThomasH99:
Yes, I suspected that might be the case, but I also noticed that at least some implementations of `initComponentImpl()` do check if `getComponentForm` returns null. And adding the check stopped my app from crashing :-) If the cause is an EDT violation, I don't think it is coming from my code. 

Unfortunately, the stack trace does not involve any of my code, so getting to the bottom of is not straightforward.

 I have a suspicion that it might be related to the animation machinery (maybe the needs which triggered stuff like `revalidateWithAnimationSafety()` and `Form.getChildrenAsList(()`... maybe `animateLayout()`/`animateHierarchy()` also need AnimationSafety versions?).

codenameone:
Yes that's true. I'll accept the PR although it's too late for today's release it will be in the next update.

PR #3131, opened 2020-06-02
Added downloadUrlSafely method and helper code

jsfan3:
After the discussion on Stack Overflow https://stackoverflow.com/a/62137379, I made many changes to the code to make it suitable to be integrated into Codename One. I also followed your suggestions.

Specifically:

1. The target file can be in both the Storage and FileSystemStorage, this difference is automatically managed.

2. Use a separate thread for merging files.

3. Callbacks are still executed in the EDT.

4. Use a unique string in the name of partial downloads to reduce problems with concurrent downloads.

5. Save partial downloads in a special folder in the FileSystemStorage to keep them more orderly.

6. Verified that the size of the output file is requested to the server only once.

7. Fixed an important error in the measurement of the downloaded data, which caused download corruption. Now the downloads are intact and even faster.

8. Added full code usage example.

### DIFF
--- a/CodenameOne/src/com/codename1/io/Util.java
+++ b/CodenameOne/src/com/codename1/io/Util.java
@@ -2087,9 +2087,11 @@ public class Util {
      * download as soon as network conditions allow and in a completely
      * transparent way for the user; note that in the global network error
      * handling, there must be an automatic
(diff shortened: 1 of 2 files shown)


codenameone:
Thanks!

PR #3161, opened 2020-07-15
Date spinner3 d keep time of day

ThomasH99:
WIthout this fix, DateSpinner3D may alter a Date even when no edits are made, simply because when getting the Date, it uses the current time of day, instead of the time of day of the original Date provide.


shai-almog:
Thanks!

PR #3162, opened 2020-07-15
Add materiel icons to span label span button

ThomasH99:
Added additional methods needed to make it possible to use a SpanLabel or a SpanButton as drop-in replacement for a Label or Button (e.g. when text becomes too long and needs to be wrapped)


ThomasH99:
I just tried compiling for Android, but I get a crash on the device, NoSuchMethodError for setMaterialIcon for SpanButton. Since my app compiles locally with latest CN1 sources including this method, I guess this means the changes didn't get pushed to the server (I thought that happened every Friday)?

shai-almog:
I'm not sure why @shannah merged this. The change didn't impact any files. See the tab for "files changed". 
I think one of your merges/commits deleted the changes you made.

ThomasH99:
Thanks. Seems I forgot to commit the changes on my own branch - I'm not at ease with Github :-( . I've created a new PR that hopefully is OK

PR #3210, opened 2020-08-10
Overloaded fetchAsPropertyList and getAsPropertyList to allow specifying the root element's key

diamondobama:


shai-almog:
Thanks!

PR #3215, opened 2020-08-11
Image.exifRotation rewriting

jsfan3:
Improvements:

* correct handling of EXIF Orientation Tag 8: missing image flipping (in addition to rotation);
* significant speed increase (tested on Samsung Galaxy A20);
* introduction of a new API that does not require saving the rotated image in FileSystemStorage, in case this is not necessary;
* small javadoc corrections.

In this code rewriting, I made sure that the returned image is still an EncodedImage as in the previous code: it seems that this is a common practice in many Codename One APIs.

I tested again all possible EXIF Orientation Tags to make sure that the code works as desired.

As far as the speed increase is concerned, the advantage is considerable, as discussed here: https://stackoverflow.com/questions/63332389/rotate-an-image-in-android-instantly-or-almost From my tests, however, it seems that this advantage is only possible if an image is rotated without preserving its transparency, so in this new version of the code transparency is not maintained. Perhaps this is due to the internal implementation of image rotation, however in the case of photographs (which is our use case) there is no transparency, so the problem does not exist.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Image.java
+++ b/CodenameOne/src/com/codename1/ui/Image.java
@@ -36,6 +36,7 @@ import com.codename1.ui.util.ImageIO;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.HashMap;
 
 /**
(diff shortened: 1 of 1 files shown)


shai-almog:
Thanks!

PR #3229, opened 2020-08-15
OFFLINE_FIRST should work like SMART when there's no cached data.

diamondobama:


shai-almog:
Thanks!
BTW shouldn't this be documented somewhere in the OFFLINE_FIRST JavaDoc?

PR #3238, opened 2020-08-24
Added support for using another font than Material for icons defined directly for a Command

ThomasH99:
It is possible to use another icon font than Material for icons defined for `Label`, `Button` etc. 

When the icon is defined as part of a `Command` (like in `Command.create(String name, Image icon, final ActionListener ev)`) a similar option is needed,.

In particular for `Commands` added to the `Toolbar` where there is no easy way to get to the `Button` that the `Toolbar` automatically creates for added `Commands`.


ThomasH99:
I meant:`Command.create*Material*(String name, char icon, final ActionListener ev)`

shai-almog:
Thanks!

Issue #3152, opened 2020-07-07
RFE: Add latest material icons


ThomasH99:
I was looking for some icons for a specific case and found a few new ones that Google seems to have added since the update mentioned here https://github.com/codenameone/CodenameOne/issues/2421. 

E.g. north, north_east etc And maybe more. If ever @jsfan3 wants to run the comparison script again... :-)

jsfan3:
You did very well to put a link to the old discussion, otherwise it would have been hard to remember what we did.

Based on the previous discussion, I assume that Codename One includes version 47 of Material Fonts, whose current version is 53, as you can see from this file:
https://fonts.googleapis.com/icon?family=Material+Icons

This is the link to the updated TTF:
https://fonts.gstatic.com/s/materialicons/v53/flUhRq6tzZclQEJ-Vdg-IuiaDsNZ.ttf

According to https://fontdrop.info/, it includes 1537 ligatures.

This is the link to the font currently included in Codename One:
https://github.com/codenameone/CodenameOne/blob/master/CodenameOne/src/material-design-font.ttf

According to https://fontdrop.info/, it includes 1278 ligatures.

There should therefore be 1537-1278=259 more ligatures, or 259 more material icons.

At the moment, however, I have not found a way to make the comparison in an automatic or semi-automatic way.

jsfan3:
@ThomasH99 You can also try to find a method to get the list of ligatures included in a ttf, that is the list of material icons and their code.  Once you have obtained the lists, the "diff" command is sufficient for comparison.

beazl:
I have created a diff to the latest font but the codepoints reference has been removed from the repo here https://github.com/google/material-design-icons so it's not much use.  There are several issues already logged asking for the codepoints file to be added back and assuming this is done, it will be pretty easy to create a useful diff.

beazl:
They've added the codepoints file and I've got the updates (284 new, 2 changed) - what's the most useful way to provide this?

jsfan3:
@beazl-peter Thank you!

ThomasH99:
And a great thank you from me as well! I looked into this but never got around to it (got pulled into designing my own icons in fontforge instead)

beazl-peter:
@jsfan3 @ThomasH99 happy to help - once they released the codepoints it became much easier!

Here is the full list of changes for reference:

# 290 new
MATERIAL_5G = '\uEF38';
MATERIAL_6_FT_APART = '\uF21E';
[...]

PR #3259, opened 2020-09-08
Update to latest material design font

beazl-peter:
#3152 updated from https://github.com/google/material-design-icons and https://material.io/resources/icons/

284 icons added
MATERIAL_GOAT removed as the previous codepoint was invalid


shai-almog:
Thanks!

PR #3273, opened 2020-09-26
Added two new validation constraints ExistIn and NotExistIn.

diamondobama:


shai-almog:
Won't it make more sense to have a more generic `NotConstraint` that will invert validation instead of `NotExistInConstraint`? 

So if one of its child constraints is true it will be false. E.g. 

```java
public class NotConstraint implements Constraint {
     // ...
     public NotConstraint(Constraint... children) {
         //...
     }

     @Override
     public boolean isValid(Object value) {
        for(Constraint c : constraints) {
            if(c.isValid(value)) return false;
        }
        return true;
    }
}
```

diamondobama:
Done

shai-almog:
Thanks!

PR #3276, opened 2020-09-28
Allowed setting InfiniteProgress material design sizes in theme constant

diamondobama:


shai-almog:
Thanks!

ThomasH99:
Nice one! I had the same need but hacked a solution. I’ll update my code to use this. Thanks

PR #3285, opened 2020-10-11
Update Component.java

sergeyCodenameOne:
We initialize the X and Y variables with getAbsoluteY( ) and getAbsoluteX( ) methods. so no mether how much we scroll the page the difference between the Y of the Component and the Y of the container will always be the same.

So if we change the initialization to be:

int x = getAbsoluteX() + scrollX;
int y = getAbsoluteY() + scrollY;

It will fix the problem.


shai-almog:
@shannah please review

shannah (approved):

PR #3318, opened 2020-12-01
Allowed for clearing TextFields on Android without having to toggle the keyboard down and up each time

javieranton-zz:


shai-almog:
Thanks.

This seems like a partial work. Why have something like this that requires the invocation of native code and don't have the full change as part of the PR?

Also the if/else is missing the curly braces which we use on single lines too.

javieranton-zz:
Yes, CN1 could have a convenience method that specifically does this behind the scenes (points 1,2,3,4) to clear a textfield without toggling the keyboard on Android. My PR only focuses in actually enabling this possibility with the introduction of the new flag.
I will create a new PR

PR #3348, opened 2021-01-17
Aligned iOS local fg notification behavior with Android (localNotificationReceived only triigers when tapped, not when shown)

javieranton-zz:
As of today, after @shannah latest changes that enabled iOS foreground notifications, iOS triggers a call to localNotificationReceived as soon as a local notification is received

Android however only does this when the local foreground notification is tapped, which makes more sense

This change makes iOS behave same as Android

### DIFF
--- a/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
+++ b/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
@@ -394,8 +394,6 @@ CN1BackgroundFetchBlockType cn1UIBackgroundFetchResultCompletionHandler = 0;
         {
             CN1Log(@"Received local notification while running: %@", notification);
 
-            NSString* alertValue = [notification.request.content.userInfo valueForKey:@"__ios_id__"];
-            com_codename1_impl_ios_IOSImplementation_localNotificationReceived___java_lang_String(CN1_THREAD_GET_STATE_PASS_ARG fromNSString(CN1_THREAD_GET_STATE_PASS_ARG alertValue));
             if (completionHandler != nil) {
                 if ([notification.request.content.userInfo valueForKey:@"foreground"] != NULL) {
                     completionHandler(UNNotificationPresentationOptionAlert);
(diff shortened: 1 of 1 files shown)


Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
```diff
@@ -589,11 +594,6 @@ -(void)application:(UIApplication*)application didChangeStatusBarFrame:(CGRect)o
 
 - (void)application:(UIApplication*)application didReceiveLocalNotification:(UILocalNotification*)notification {
     CN1Log(@"Received local notification while running: %@", notification);
```
shannah, line 596: `application:didReceiveLocalNotification:` is still required for iOS <10.    `userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:` is only available in iOS >=10.

Hence, this change will break local notifications for iOS<10.

What device are you testing on?

javieranton-zz: You're absolutely right, I had a feeling it would be the case but couldn't test because I only have iOS 10+ devices (iPad 4 & iPhone 6s)

javieranton-zz (commented):
You're absolutely right, I had a feeling it would be the case but couldn't test because I only have iOS 10+ devices (iPad 4 & iPhone 6s)

javieranton-zz:
Hi, I must have missed your 13 day old comment. Apologies, I was convinced there wasn't a reply. To answer your question: yes, I have tested this extensively. Like thousands of times, I also have released builds with these modifications and tested in production. I always end up having to edit the delegate so all merging this will do is just save me having to do that. But I thought it would make sense for everyone too if it behaved this way

I have tested on iOS 10 and iOS 13, so if you can test on lower versions that would be great too (I don't have any device with lower than iOS 10)

In older iOS versions, local notifs will still go straight to localNotifReceived, but I guess that's fine

shai-almog:
This will have to wait to next week as we're in code freeze for 7.0. Please ping us again in the middle of next week and we'll try to merge it assuming @shannah has no objections.

javieranton-zz:
> This will have to wait to next week as we're in code freeze for 7.0. Please ping us again in the middle of next week and we'll try to merge it assuming @shannah has no objections.

No worries Shai, I'll try to remember

javieranton-zz:
Ping. TBH, I'm not sure this should be merged. There is a clear user benefit but unless someone can fully test this with CN1's push it might break things. I implement my own push so no matter how much I test it won't cover all scenarios. OK to close

PR #3355, opened 2021-02-02
Support set for externizable

ThomasH99:
Current a Set (as a Collection) is externalized as a list and internalized as an ArrayList which means it won't work correctly after internalization.


shai-almog:
Please remind us about this next week. We're in a code freeze for 7.0 and can't merge additional features right now. This shouldn't be a problem right after the release.

PR #3384, opened 2021-03-31
Finger tracking in form

ThomasH99:
Added support for super-imposing images below the pointer (fingers). I cannot test it on the device (can't build for device with locally modified sources) so I hope you'll review the code before accepting.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Form.java
+++ b/CodenameOne/src/com/codename1/ui/Form.java
@@ -4816,7 +4816,7 @@ public class Form extends Container {
 
     /**
      * called by Forms, Dialogs, Menus etc created from this Form to
-     * automatically set the PointerTracking
+     * automatically set the PointerTracking 
      */
     private void initPointerTracking() {
         if (Display.getInstance().getCurrent() != null) {


shai-almog:
Thanks!

This code won't compile on our servers since it uses Java 8 lambda syntax. The code should be limited to Java 5 syntax for core code.

I would also refactor this so the glass pane is a separate class and the API should only be exposed on that class. So we'd have a new class "PointerTracker" through which this API will be installed. It can be in the same package and use package friendly APIs into Form but that way we'd keep the API changes segregated and clear. Does that make sense?

ThomasH99:
Thanks for looking at this, but not sure I get how you want the code structure. Can you point to an example I could use for inspiration?

shai-almog:
I'm not sure if there's something similar in the code I can point you to. My goal is to keep the impact on Form.java to a minimum and ideally make it unaware of most of this change.  The logic behind this is two fold:

- Easier to change later on and extend
- Smaller API in the core for functionality that isn't critical

To do that you can introduce a new class e.g. `PointerHighlightGlassPane`. It would include some of the code from `setPointerTracking` and most of the other code/APIs. It would essentially install itself on a `Form` instead of invoking an API on Form you would invoke it on that class.

PR #3509, opened 2021-10-07
Fix Maven configurations so "mvn install" can be run without errors

Double-A-92:
The maven build was just not working for me, when trying to run it via IntelliJ on JKD11. 
This were the changes I had to make to make everything run without errors or warnings.


Double-A-92:
> Thanks. Not to be picky. But can you squash the commits in this PR so it is easier for me to see exactly what the merge will do?

Ok did it. I basically just grouped some things in the pom file.

PR #3531, opened 2021-11-21
Added Tabs.setSwipeOnYAxis(...) method to set the swipe on Y-Axis.

jsfan3:
This addition follows our discussion:
https://new.reddit.com/r/cn1/comments/quq7yo/realize_a_set_of_containers_that_are_browsable/

To test the code, you can try this example:
https://gist.github.com/jsfan3/92c8b852157c4196e11825c330641dae

The changes I've made should have no impact on the existing code, or so I think. I have followed as much as possible the logic of the pre-existing code.

For the moment, I have only tried in the Simulator.

I have seen a minor pre-existing bug; it does not depend on my modifications. It occurs only if the tabs have a margin: in such a case, for a fraction of a second, the Motion moves the incoming tab entirely to the left during the swipe (thus ignoring the margin). This bug is almost undetectable in a move along the X-axis. However, if the swipe is along the Y-axis, the bug becomes visible. It's a minor problem that I don't know how to fix; for the moment, the workaround is to have a zero margin (as in the above-linked sample code).

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Tabs.java
+++ b/CodenameOne/src/com/codename1/ui/Tabs.java
@@ -60,6 +60,9 @@ import com.codename1.ui.util.EventDispatcher;
  * <script src="https://gist.github.com/codenameone/ba27124a0a25e685b123.js"></script>
  * <img src="https://www.codenameone.com/img/developer-guide/components-tabs.png" alt="Simple usage of Tabs" />
  * 
+ * The <code>Tabs</code> allows swiping on the X-axis (by default) but also on the Y-axis:
+ * <script src="https://gist.github.com/jsfan3/67074c6684b0ee711c6f6d950cdefb57.js"></script>
+ * 
  * <p>A common use case for {@code Tabs} is the iOS carousel UI where dots are drawn at the bottom of the 
  * form and swiping is used to move between pages:</p>
  * <script src="https://gist.github.com/codenameone/e981c3f91f98f1515987.js"></script>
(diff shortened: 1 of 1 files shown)


shai-almog:
I have two small nitpicks with the code. 

The variable is swipeX while the method is swipeY. I suggest making the method swipe X and setting it to false would do what you want.

This would also fix the second problem. 

Right now you can't toggle this back and forth because of the `if` statement in that method.  By just exposing the variable this should be solved.

You would also need to add an isSwipeX method for consistency.

jsfan3:
I made the changes you asked for and updated the javadocs.

The new sample code is: https://gist.github.com/jsfan3/67074c6684b0ee711c6f6d950cdefb57

There are two minor bugs, and I don't know how to deal with them. These two bugs don't affect my use case; however, they might affect other users.

The first one concerns the use of buttons to change tabs (so hideTabs was not invoked): the swipe effect when changing tabs happens only if the swipe is on the X-axis; otherwise, the tab change is immediate. If the developer uses hideTabs, the problem does not arise.

The second bug results from the possibility of invoking setSwipeOnXAxis as many times as you want after the Tabs component is displayed. The swipe effect does not happen correctly on the currently displayed tab, but the problem disappears at the first tab change. If the developer invokes setSwipeOnXAxis only once and before displaying Tabs, the problem does not arise.

How do you suggest proceeding?

jsfan3:
Right now, I'm working on the two problems I listed in my last comment.

One of them I solved: basically, I forgot to use the variable "swipeOnXAxis" also inside the methods "animate()" and "setSelectedIndex(...)".

I hope to solve soon also the other problem. Then I'll send another commit when I'm all done.

jsfan3:
I just solved the other problem as well. I forgot that the "contentPane" of Tabs changes its size and internal tab positioning when we invoke "setSwipeOnXAxis" with a different value than the current one, so that method has to call internally "revalidateLater".

Before submitting the new solving commit, I want to add a short demo video to the Javadoc.

jsfan3:
I finished. I don't know what the failed check means.

shai-almog:
Thanks!

PR #3551, opened 2022-02-04
Fix audio playback within Simulator.

LevittS:


shannah:
Thanks for the pull request.  Can you provide a test case that this fixes for you.  I seem to recall that using https:// (rather than a custom scheme) solved a lot of issues with webpage loading in CEF.  My original implementation used the custom scheme (like you change it to here) but I had to change it because of problems (I don't remember specifics).
If you can provide a test case, it will help me work with your pull request to fix it for all cases.

LevittS:
I couldn't get any audio to work without changing something - just single createMedia calls followed by play regardless of the source of the InputStream weren't working for me - I was seeing network errors in the CEF browser console (a bit like cn1app wasn't a valid hostname).
I'm away from my machine at the moment, and could provide some actual code later, but it was basically the same as the example code I could find.

PR #3622, opened 2022-08-28
Added support to SwipeBack for return null destination form as a means to de-activate swipeBack if conditions to exit Form are not met

ThomasH99:


shai-almog:
Thanks!

Issue #3646, opened 2022-10-13
Maven Plugin and TestRunner do not support JUnit XML properly


Rocketeer007:
**Describe the bug**
When running a build via Maven, there is no way to get JUnit-formatted test reports. This makes it difficult to include test results from CN1 in a Jenkins build.

Properly, these are actualy Surefire format, since Surefire is what generates the reports for JUnit tests in a typical Maven build.

Additionally, running TestRunner directly rejects the `-junitXML` parameter, even though it's defined in the code.

**To Reproduce**
1. Run `mvn -Pjavase package` and observe that CN1 tests are run
2. Check for any surefire-format XML files being generated; there are none.

1. Run `java -cp... com.codename1.impl.javase.TestRunner com.my.Project -junitXML`
2. Observe failure with output as below
```
Unrecognized argument: -junitXML
Usage: TestRunner mainClass [-testCases testCase1,testCase2...] [-skins skin1,skin2...] [-quiet] [-cleanMode] [-junitXML][-stopOnFail]

mainClass - the main application class which is being tested, this is the full name for the lifecycle class.
testCases - optional set of test cases to run using the full package/class name syntax and comma delimited. If ommitted all test cases will be executed.
skins - the skins on which the test cases should be executed. If ommitted the default simulator skin is used.
quietMode - when set the skin running the tests will not appear and the tests will be executed in the background
cleanMode - clean mode executes every test in complete isolation from the previous test restarting the Classloadercompletely. Clean mode can't be used on the device so its only useful for debugging
stopOnFail - indicates that execution should stop the moment a failure occured
junitXML - output is written in junit XML format
```

**Expected behavior**
Ideally, running a Maven build should generate Surefire-format reports that can be used in a CI system.

Runnning the `TestRunner` directly should not reject a parameter that's included in the Usage.

**Additional context**
I am currently working on a PR to adress these issues, but wanted to have an issue logged for reference.

PR #3647, opened 2022-10-13
Issue #3646: Generate Surefire-compatible XML when running tests via Maven

Rocketeer007:
* Fixed `TestRunner` to correctly support the `-junitXML` parameter
* Implemented generation of XML file with test report on completion of a test suite
* Added a `testName` paramter to `TestReporting.testExecutionFinished` as the existing implementation for the test name was generating an NPE
* Changed `JUnitXMLReporting` to also use the default reporting logic (we can see output in Maven logs as well as in generated XML)
* Improved `TestExecuter` so that test exceptions are logged correctly
* Changed `RunTestMojo` to enable `-junitXML` when calling `TestRunner`, and generate a `"cn1-reports"` directory for the output

### DIFF
--- a/CodenameOne/src/com/codename1/testing/DeviceRunner.java
+++ b/CodenameOne/src/com/codename1/testing/DeviceRunner.java
@@ -79,9 +79,9 @@ public abstract class DeviceRunner {
         } catch(IOException err) {
             TestReporting.getInstance().logException(err);
         }
-        TestReporting.getInstance().testExecutionFinished();
+        TestReporting.getInstance().testExecutionFinished(getClass().getName());
         if(failedTests > 0) {
-            Log.p("Test execution finished, some failed tests occured. Passed: " + passedTests + " tests. Failed: " + failedTests + " tests.");
+            Log.p("Test execution finished, some failed tests occurred. Passed: " + passedTests + " tests. Failed: " + failedTests + " tests.");
         } else {
             Log.p("All tests passed. Total " + passedTests + " tests passed");
         }
@@ -128,7 +128,7 @@ public abstract class DeviceRunner {
                             TestReporting.getInstance().finishedTestCase(t, false);
                         }
                     }
-                };
+                }
                 RunTestImpl runTest = new RunTestImpl();
                 if (t.shouldExecuteOnEDT() && !CN.isEdt()) {
                     CN.callSeriallyAndWait(runTest);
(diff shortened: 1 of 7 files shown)


shai-almog:
Do you need help understanding the build errors?

https://github.com/codenameone/CodenameOne/actions/runs/3240963162/jobs/5312265710

Rocketeer007:
I have to confess, I'm really confused by those build errors. 

I'm assuming the root cause is that the build job is using an older version of Java than I'm using locally, which doesn't include some of the classes and methods I'm using.  However, the build job specifies Java 8, and I'm using Java 8... and `PrintWriter` has been around since JDK1.1.

I've taken a wild guess, and replaced the wildcard import with explicit imports, and swapped out the more complex uses of `Throwable` to see if that makes a difference...

Rocketeer007:
Okay, seen the new failures, which suggest the build is fine, but the VM is not happy passing around `UnitTest` instances from different classloaders.  This makes sense - one of the changes I had to make was to prevent separate instances of `TestRunner` from being created by each classloader, as otherwise we only got half the information about the tests.

Oddly, actually using the new classes in Maven and in TestRunner did work, so I'm not 100% sure why it fails in the build infrastructure... so the first thing I'm going to do is figure out how to run these tests locally and reproduce the problem.

My proposed solution will be to modify the `TestReporting` API to take the test name as a String, rather than the `UnitTest` object itself - I originally had that as part of this PR, but ended up reverting it as "too wide-reaching".

shai-almog:
This looks good to me. Just one small caveat.  You changed a couple of public method signatures e.g.:

```java
public void testExecutionFinished(String testSuiteName) {
}
```

This might be OK but in theory someone might have subclassed and overriden these APIs. I think the right thing to do here would be something like:


```java
public void testExecutionFinished(String testSuiteName) {
    testExecutionFinished();
}

/**
 * @deprecated Please use #testExecutionFinished(String)
 */
public void testExecutionFinished() {
}
```

Does that make sense?

@shannah what do you think?

Rocketeer007:
Added deprecated methods as you suggest. If anyone has overridden the `startingTestCase` or `endingTestCase` methods, they'll have to make changes to restore functionality to their reporting mechanism, but at least it will still compile, and the deprecation warnings will tell them where to look.

Anyone _using_ CodenameOne should be unaffected, since I can't see any way to successfully inject a different reporting tool while running tests that use this framework (calling `TestReporting.setInstance` during a test would have changed the reporting tool used in that test, but since the tests used a different classloader than the test framework, it would have been quite a mixture of different tools depending on what exceptions occurred and when.

Anyone _extending_ CodenameOne and supplying a new `TestReporting` instance will have to make changes for it to work correctly.

shai-almog:
Great. Thanks for all the hard work!

shai-almog:
@all-contributors please add @Rocketeer007 for code

PR #3663, opened 2022-11-28
Scrolling improvements

### LINKED ISSUE #3657: Android Port Doesn't Handle Mouse Events Properly, Specifically Scroll Wheel
Looking at the code we will probably need to add a listener to mouse events [here](https://github.com/codenameone/CodenameOne/blob/master/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java#L4232) and some handling similar to [this](https://github.com/codenameone/CodenameOne/blob/master/Ports/Android/src/com/codename1/impl/android/CodenameOneView.java#L443).

Android docs have some discussion about the types of events available in newer versions of Android that should include those gestures [here](https://developer.android.com/guide/topics/large-screens/input-compatibility-large-screens#navigation).

Tianfeng-Chen:
Fix #3657 
- Use setOnGenericMotionListener to handle scrolling event
- The logic is detecting input from source ***InputDevice.SOURCE_CLASS_POINTER*** and verify if this event is a motion event with a type of ***MotionEvent.ACTION_SCROLL***
- Get current pointers' X axis and Y axis value
- Using ***event.getAxisValue(MotionEvent.AXIS_VSCROLL)*** to get the moving direction indicator (this can get a value between -1 and 1)
- Use the current Y axis value to calculate the location that the scene should be scrolled to
- Keep the same X axis value and utilize ***scrollTo()*** method to scroll the scene
- Add constrains to verify the desired scrolling location is not off the screen.

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
+++ b/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
@@ -4245,15 +4245,21 @@ public class AndroidImplementation extends CodenameOneImplementation implements 
                                     if (AndroidImplementation.this.myView == null) {
                                         return false;
                                     }
-                                    if (me.isFromSource(InputDevice.SOURCE_CLASS_POINTER)) {
-                                        switch (me.getAction()) {
-                                            case MotionEvent.ACTION_SCROLL:
-                                                int currentX = (int) me.getX();
-                                                int scrollDistance = (int) me.getAxisValue(MotionEvent.AXIS_VSCROLL);
-                                                int scrollToLocation = (int) (scrollDistance * 5d);
-                                                myView.getAndroidView().scrollTo(currentX, scrollToLocation);
-                                                return true;
+                                    if (me.isFromSource(InputDevice.SOURCE_CLASS_POINTER) &&
+                                            me.getAction() == MotionEvent.ACTION_SCROLL) {
+                                        double currentX = me.getX();
+                                        double currentY = me.getY();
+                                        int maxHeight = myView.getViewHeight();
+                                        double scrollOffSet = me.getAxisValue(MotionEvent.AXIS_VSCROLL);
+                                        int scrollToLocation = (int) (scrollOffSet * 5d + currentY);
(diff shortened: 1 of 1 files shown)


shai-almog:
@shannah please review

shannah:
This PR only appears to apply to peer components, not the general app. 
The most obvious native component to test this on is the webview, but it has its own scrolling functionality, that would need to be tested thoroughly to make sure it doesn't conflict with this.

So, before merging this I need to:
1. Create a test app with no native peer components, to ensure no regressions.
2. Create a test app with a webview, and test with and without this change, and with and without scrolling enabled on the webview to make sure that it doesn't introduce regressions, and verify whether it improves anything.
3. Create a test app with a non-webview native peer component and see how this affects the scrolling on that component.

Also, I sense that this was not the intention of the PR.  The title of the PR and the associated issue make is sound like this was intended to add scroll-wheel support to CN1 android apps in general.  I don't think it does that.
Probably this sort of thing would need to be added in the CodenameOneView class:
https://github.com/codenameone/CodenameOne/blob/master/Ports/Android/src/com/codename1/impl/android/CodenameOneView.java#L372

And possibly the AsyncView also.

Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
```diff
@@ -4235,6 +4235,36 @@ public boolean onTouch(View v, MotionEvent me) {
                                     return myView.getAndroidView().onTouchEvent(me);
                                 }
                             });
+                            /**
+                             * Mouse Event Listener used to handle mouse wheel events
+                             * The mouse wheel can be used to scroll the page up or down
+                             */
+                            layoutWrapper.setOnGenericMotionListener(new View.OnGenericMotionListener() {
```
shannah, line 4242: This adds a motion listener to a native peer component. E.g. a webview or other native component.  Not the main view.  Probably we need to add similar in CodenameOneView if we want this to be for general app.

shannah (changes requested):

PR #3685, opened 2023-03-10
removed and added some logs

snur206:


shai-almog:
I assume this was submitted by mistake, right?

PR #3736, opened 2023-09-25
Change stringbuffer.java/stringbuilder.java

SufiaRashid:
Made changes to methods within StringBuilder.Java and StringBuffer.Java based on issue completion request

image


shai-almog:
Can you remind me the issue that this relates to?

Notice you commented out `subSequence` which seems ill-advised as it would break existing APIs.

PR #3740, opened 2023-09-30
Update .all-contributorsrc

fardeen2k:


shai-almog:
It seems that this was submitted by mistake. Right?

Issue #3702, opened 2023-05-04
Vibrate doesn't work in android and ios


DurankGts:
Display.getInstance().vibrate(1000);

why this code doesn't work?

DurankGts:
I just add this to codename1.arg.android.xpermissions = <uses-permission android:name="android.permission.VIBRATE" /> and it actually work in android 11 but not in ios.

Any support about this?

shai-almog:
It seems the code [here](https://github.com/codenameone/CodenameOne/blob/master/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java#LL2605C70-L2605C86) should be updated based on [this](https://stackoverflow.com/questions/68466095/vibrator-service-string-is-deprecated-for-api-31).

On iOS there's no guarantee vibrate will work as mentioned [here](https://developer.apple.com/documentation/audiotoolbox/1405248-audioservicesplaysystemsound?language=objc). Relevant paragraph:

> On some iOS devices, you can pass the [kSystemSoundID_Vibrate](https://developer.apple.com/documentation/audiotoolbox/1618202-alert_sound_identifiers/ksystemsoundid_vibrate?language=objc) constant to invoke vibration. On other iOS devices, calling this function with that constant does nothing.

This applies to [this code](https://github.com/codenameone/CodenameOne/blob/master/Ports/iOSPort/nativeSources/CodenameOne_GLViewController.m#LL1290C5-L1290C57).

So this should only be fixed for Android as far as I can tell.

regulas1:
Hi There May I take This Up ? I am new to Open Source so I want to try this one. 
Please let me know if it good for me to work with?

shai-almog:
Sure @regulas1 
assigned the issue to you.

regulas1:
Hi @shai-almog  can you check if the code below is the correct fix ?
`               //Replaced this
                v = (Vibrator) getContext().getSystemService(Context.VIBRATOR_SERVICE);
                // With the code below
                val v: Vibrator = (getSystemService(Context.VIBRATOR_SERVICE) as? VibratorManager)?.defaultVibrator
                        ?: (getSystemService(Context.VIBRATOR_SERVICE) as Vibrator)`

shai-almog:
The second code is in Kotlin, not Java. So no it isn't correct. You need to adapt the Kotlin syntax to Java.
We only support Java 6 level syntax at the moment so you can't even use current syntax levels.

regulas1:
[...]

PR #3755, opened 2023-10-28
Implemented API 31+ vibration service code to replace depreciated code

u7484052:
Added code that fixes the following issue:
https://github.com/codenameone/CodenameOne/issues/3702

I added retro-compatibility as well if needed.


shai-almog:
Can you remove the extra files from the commit? 
There are 12 project files that shouldn't be a part of the commit.

shai-almog:
Also please fix the compilations errors seen in the CI run. I'm guessing there's a missing import and you should use numeric values instead of constants for backward compatibility with older compilers.

u7484052:
Yeah sorry, I have fixed the importing error and removed the unnecessary files from the PR

shai-almog:
This is still failing. I'm guessing it requires a fix to the library in the classpath but I'm not sure if this is something you can do. If not I'll have to look into it.

PR #3757, opened 2023-10-29
Feature/plugin support

shannah:
- Added support for plugins which can override core functionality in Codename One.
- Added support for `OpenGallery` and `IsGalleryTypeSupported` plugins to allow us to write plugins that override the `Display.openGallery()`, `Display.openImageGallery()`, and `Display.isGalleryTypeSupported()` methods via plugins.
- Added new `core-unittests` maven module that runs pure unit tests on the Codename One core using JUnit and Mockito.
- Added step to the pr.yml Github workflow to run the core-unittests


CodenameOne/src/com/codename1/plugin/PluginSupport.java
```diff
@@ -0,0 +1,106 @@
...
+    private static PluginSupport instance;
+
+    private static final Object instanceLock = new Object();
+
+    /**
+     * Gets shared instance of PluginSupport class.
+     *
+     * @return Singleton instance
+     */
+    public static PluginSupport getInstance() {
```
shannah, line 48: TODO: Get rid of this.  Instead, just add a `getPluginSupport()` method to `Display` and use that as the "singleton" instance.

Issue #3761, opened 2023-11-12
[Critical] iOS native theme does not adapt to the iPhone 'dynamic island' (which replaces the 'notch')


ThomasH99:
The iOS theme and SafeArea does not adapt to the iOS 'Dynamic Island'. The protected space is not high enough so a title text which is well positioned under the notch on earlier iPhones (with the notch) is now partly hidden below the dynamic island. Which is not surprising since the dynamic island descends further on the screen than the notch, so more space needs to be included in the safeArea.

This is pretty critical as the full line of iPhones now uses this concept.

shannah:
We have found the issue, and the fix will be available in the next update on Friday.
In the mean time, please try this workaround and see if it fixes the issue for you.

On all forms in your app, after you have initialized the toolbar, call 

```
Container statusBar = ComponentSelector.select("StatusBar", form).asComponent(Container.class);
if (statusBar != null) {
   statusBar.setSafeArea(true);
}
```

ddyer0:
It seems likely that this fix broke the toolbar logic for Iphone 12 with the hardwired notch.  See
https://www.reddit.com/r/cn1/comments/1b91u19/toolbars_are_broken_wrt_safe_area_logic/

My temproary fix is to set padding on the toolbar to safeArea.top() before getPreferredSize()
of the toolbar.
'''java
	public Dimension getPreferredSize()
	{	// this is the essential repair to the toolbar logic.  Include
		// the height of the safe area in the toolbar size.  Other kludgery
		// in com.codename1.ui.Container will add a margin corresponding to
		// the safe area.
		Style s = getStyle();
		s.setPaddingUnitTop(Style.UNIT_TYPE_PIXELS);
		s.setPaddingTop(getSafeAreaHeight());
	}
'''
This version would work if getSafeArea is reliable, which it is not.

PR #3763, opened 2023-11-12
fix: Fixed toolbar overlapping dynamic islands on iOS

shannah:
The toolbar will now set the status bar as a safe area, if it creates the statusbar spacer (which it does on iOS).  Otherwise the toolbar itself will be a safe area. Fixes https://github.com/codenameone/CodenameOne/issues/3761

Screenshot with this fix employed.

Before this fix:

image

After this fix:

image


shannah:
I see there are some test failures. That is due to the fact that this is pushing the position of some components down properly now to accommodate safe areas now.  I need to adjust these tests.

Issue #3793, opened 2024-03-08
"safe area" rectangle is unstable, and incorrect, for iphones


ddyer0:
AFAIK the first time display.getSafeArea() is called, the rectangle is correct, whichever orientation
the phone is in.  Subsequently, each time you flip from portrait to landscape, the safe area is
modified to a different, incorrect value, where the width and height are swapped from the current
width and height that are being assigned to the new display being constructed.

safe x=117 y=0 w=1845 h=897
safe x=0 y=117 w=960 h=1878
safe x=117 y=0 w=1962 h=876
safe x=0 y=117 w=960 h=1878
safe x=117 y=0 w=1845 h=897

ThomasH99:
I’ve experienced a similar issue several times, so I think there is a real bug Sent from my iPhoneOn 8 Mar 2024, at 08:43, ddyer0 ***@***.***> wrote:﻿
AFAIK the first time display.getSafeArea() is called, the rectangle is correct, whichever orientation
the phone is in.  Subsequently, each time you flip from portrait to landscape, the safe area is
modified to a different, incorrect value, where the width and height are swapped from the current
width and height that are being assigned to the new display being constructed.
safe x=117 y=0 w=1845 h=897
safe x=0 y=117 w=960 h=1878
safe x=117 y=0 w=1962 h=876
safe x=0 y=117 w=960 h=1878
safe x=117 y=0 w=1845 h=897

shannah:
@ddyer0 In your readout there, which outputs are "incorrect"?

ddyer0:
all of them.  tThe lines wheere x=0 correspond to holding the phone horizontally, where the notch is either lift or right.  
In general, whichever orientation the app is initially launched in, the safe area is correct, then is corrupt for all subsequent times the phone is rotated.  getDisplayWidth and getDisplayHeight are also incorrect, and incompatible with the dimensions of the form itself.

ddyer0:
''''java
/**
 * demo program for issue 3973
 */
package com.boardspace.dtest;

import com.codename1.ui.Component;
import com.codename1.ui.Container;
// issue #3645
// 
// drawing pdfs
//
//
import com.codename1.ui.Display;
import com.codename1.ui.Form;
import com.codename1.ui.Graphics;
import com.codename1.ui.Label;
import com.codename1.ui.Toolbar;
import com.codename1.ui.geom.Dimension;
import com.codename1.ui.geom.Rectangle;
[...]

PR #3794, opened 2024-03-10
fix: incorrect ios safe areas

shannah:
Ticket: https://github.com/codenameone/CodenameOne/issues/3793


ddyer0:
I haven't checked your work - but an additional problem with safe areas was that they were reported the same
if the actual safe area was on the right or on the left.

shannah:
Correct behaviour will be for safe area insets to be non-zero on both left and right when in landscape, to account for round corners.

ddyer0:
Have safe areas ever accounted for the rounded corners at the bottom, in portrait mode?
In any case, I'm more concerned about the big divot than the corners.

PR #3796, opened 2024-03-11
Use JSONParser.setUseLongsInstance and JSONParser.setUseBooleanInstance

Rocketeer007:
The current solution calls the static method `JSONParser.setUseBoolean`, which updates the default value for `useBoolean`, but does not update the current instance.
This means it will not take effect on the current instance - and it will also impact all future instances.

This solution replaces the calls to `setUseBoolean` with `setUseBooleanInstance`. It also adds `setUseLongs`, following the same pattern.


shai-almog:
Looks great. Thanks!

PR #3797, opened 2024-03-12
Use setUseLongsInstance and setUseBooleanInstance in PropertyIndex

Rocketeer007:
This fixes a subtle bug where the first use of `PropertyIndex.loadJSONList`, `fromJSON` or `loadJSON` would not have used Booleans or Longs, but subsequent requests would.  This is not particularly noticeable with Boolean, since the `Property` logic would coerce a String into a Boolean anyway - but it had the side-effect of setting a new default for all future uses of `JSONParser`, which could be problematic.


shai-almog:
Thanks. Great catch!

PR #3803, opened 2024-04-14
fix: fixed but with drawString rendering long RTL strings incorrectly…

shannah:
… on iOS


shannah:
Test app: https://gist.github.com/shannah/1d461bc7ade72ab591e2295ca004e2b7

PR #3808, opened 2024-06-11
Update Form.java

ImmediandoSrl:
update method called on InPlaceEditView for adding support to KEY TAB phyisical keyboard

### DIFF
--- a/CodenameOne/src/com/codename1/ui/Form.java
+++ b/CodenameOne/src/com/codename1/ui/Form.java
@@ -4313,7 +4313,7 @@ public class Form extends Container {
      * on the currently focused component.  Failing that it will scan the form based on Y-coord.
      * @return 
      */
-    public Component findNextFocusDown() {
+    Component findNextFocusDown() {
         if (focused != null) {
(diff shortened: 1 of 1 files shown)


shai-almog:
This pull request is problematic. It exposes internal implementation details and as far as I can tell it has no real purpose. It's used within https://github.com/codenameone/CodenameOne/pull/3809 to implement cyclic focus. That means we can't build https://github.com/codenameone/CodenameOne/pull/3809 until this is merged which I don't think we should.

Cyclic focus is already enabled in the code by default, if that isn't working then the problem is in the application logic. Notice that the code for traversing text components uses the `getTabIterator` method internally to find the next appropriate component. That method is public and as far as I can tell should be cyclic.

ImmediandoSrl:
ok but supposing i have 3 textfieds, the tab works correctly until the last one. 
I want to turn to the first one when i click tab on the last textfield. how can I do?

shai-almog:
Cyclic is just that. The focus moves in a cycle. Closing this PR and moving the discussion to the other PR.

PR #3814, opened 2024-06-13
Update InPlaceEditView.java

ImmediandoSrl:
add implementation for unknown listener

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java
+++ b/Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java
@@ -1168,9 +1168,13 @@ public class InPlaceEditView extends FrameLayout{
                 && ((TextArea) mEditText.mTextArea).getDoneListener() != null
                 && (actionCode == EditorInfo.IME_ACTION_DONE)|| actionCode == EditorInfo.IME_ACTION_SEARCH || actionCode == EditorInfo.IME_ACTION_SEND || actionCode == EditorInfo.IME_ACTION_GO) {
             ((TextArea) mEditText.mTextArea).fireDoneEvent();
-
         }
-        
+
+	if (reason == REASON_IME_ACTION
+                && ((TextArea) mEditText.mTextArea).getUnknownListener() != null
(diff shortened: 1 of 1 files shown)


shai-almog:
Unknown event doesn't make sense to me. If there's an important event we need to deal with it correctly, not throw it into the miscellaneous pile. That might seem like a reasonable hack when we start but pretty soon the miscellaneous pile will become the largest and people will depend on behavior that isn't cross platform.

The escape change is fine and makes sense. But isn't it effectively the same action as back? 

Why not use the same switch as that?

ImmediandoSrl:
i would like to intercept phyisical key ESC so when press i call a function that copy values from previous rilevation (price, facing) to new one.
doing unknown event as i do i can intercept with unknownlistener the keyCode of key pressed so overriding TextField i can switch to keyEvent and do what i prefer.

textfield.setUnknownListener(new ActionListener<ActionEvent>() {
        switch(evt.getKeyEvent()) {
             case 111: //Do something 
             case .... : 
    }
});

ImmediandoSrl:
with donelistener I no longer have references to the code of the pressed key

shai-almog:
A key event should arrive anyway in key events. This is a special case of native editing where the text is edited by the OS. You want to send something back to the Java layer which is currently not responsible for anything (since we're in editing mode). This means moving to a different thread and in the meantime the native OS might do other things (like add additional text). You're setting yourself up to disaster and this is Android only anyway.

The question is what is the "Do something"? 

By doing it this way you're creating fragile unportable code. Might as well implement a generic solution at the implementation layer that works correctly.

It might be possible to add the keycode that triggered the done event to the `ActionEvent` that is sent. 

E.g. [this method](https://github.com/codenameone/CodenameOne/blob/master/CodenameOne/src/com/codename1/ui/TextArea.java#L2199) can't be removed since it's public. But you can add another one that accepts an `ActionEvent` and pass your own action event from [here](https://github.com/codenameone/CodenameOne/blob/master/Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java#L1170).

I'm not sure if that's the best option since I don't have all the information.

PR #3815, opened 2024-06-13
Update TextArea.java

ImmediandoSrl:
add managing for unknown events on textarea


shai-almog:
Adding a whole new event system for a user specific edge case doesn't seem like the right decision to do in this case. Also, please try to have a single inclusive PR. This PR relates to https://github.com/codenameone/CodenameOne/pull/3814 and that PR can't compile without this. It makes sense to have only one.

ImmediandoSrl:
sorry, but every time i change a file GitHub create new branch so i have to create two different PR, but next time i'll create single PR

shai-almog:
You need to work locally for this to be easier. If you're working through the github site just fork the project. Make changes to your fork and commit them, then select the option to submit a PR. Alternatively, there's an option to make multiple changes on a single branch by selecting the same branch when editing.

PR #3816, opened 2024-06-16
KeyCode on TextArea DoneEvent

ImmediandoSrl:
PR relates to https://github.com/codenameone/CodenameOne/pull/3814

Added the keycode that triggered the done event to the ActionEvent that is sent.


shai-almog:
You can't change the signature of the `fireDoneEvent()` method. That's a public API, you will be breaking compatibility with older versions.  You can only add another method with the same name and the additional parameter. The old method can invoke the new method to prevent code duplication but you can't break the code of everyone who relied on `fireDoneEvent()`.

Other than that looks good. If I would nitpick I would revert the change to`ActionEvent.java` since it's just indentation. But that's not a big deal.

shai-almog:
Also notice the compilation error from CI:

```
    [javac] Compiling 636 source files to /home/runner/work/CodenameOne/CodenameOne/CodenameOne/build/classes
    [javac] /home/runner/work/CodenameOne/CodenameOne/CodenameOne/src/com/codename1/ui/TextArea.java:2205: error: local variable keyEvent is accessed from within inner class; needs to be declared final
    [javac]                         fireDoneEvent(keyEvent);

```

ImmediandoSrl:
yes sorry for ActionEvent, i was editing the class from tablet yesterday and i didn't find the revert button for ActionEvent class.

Regarding the TextArea, I changed the fireDoneEvent because, looking in the code, the only call was in the InPlaceEditView, so changing them would not have created incompatibility problems. I still updated the class as you asked

shai-almog:
Thanks.

Once a method is public in a framework anyone anywhere can rely on that. Changing a public method would mean breaking code you don't know about. For future PRs please try to include a JavaDoc comment for every method otherwise it looks weird in the [JavaDoc](https://www.codenameone.com/javadoc/).

PR #3819, opened 2024-06-25
Add client property to textarea

ImmediandoSrl:
Good morning, 

i added a client property to textarea to set background color to TextField on focus


shai-almog:
The background color in the case of focus should be the selected color from the style object. If that behavior doesn't happen then that's a bug. The somewhat confusing behavior occurs when touch is involved. Focus is only used for non-touch devices. Once you touch the device focus vanishes since it's no longer applicable, it is only shown when you touch a component.

If you use a physical key on the device the focus should then return. Notice that when the input invokes `getStyle()` it is implicitly asking for the right `Style` object at this particular time. That means the style should be the focused `Style` object if focus should be used.

PR #3822, opened 2024-07-15
Focus on CheckBox

ImmediandoSrl:
Good morning, i'm trying to implement focus on checkbox component with TAB on physical keyboard.
Debugging code i see that when i press TAB on my Device with physical keyboard from TextField to CheckBox it doesn't work.

It only works switching from a TextField to another, because of this code on Runnable onEditorAction in InPlaceEditView class (line 1233)
            if (next != null && next instanceof TextArea) {
                nextTextArea = (TextArea) next;
            }

Adding fNext.requestFocus(); it works fine. is possible to publish this request or suggest me another way to do that?


shai-almog:
As far as I can read the change you're requesting focus to the next component while processing an action in the currently editing component. This can have multiple edge cases and problems that impact everyone and also doesn't seem like the right thing. A change like this should be handled for the tab key only, not for everything.

As a side note the indention for both lines modified is inconsistent with the rest of the code making it harder to read.

ImmediandoSrl:
Good morning, I modified my pull request now the component will only request focus if the TAB key was pressed

shai-almog:
Thanks.
I merged this. But notice that the indentation is still broken. You can see it in the diff here:
https://github.com/codenameone/CodenameOne/pull/3822/files/81d2dd9623676ab20c62105b71bf06286f4b9e41..0851bd3ab92e6504dc80d8086ae244e2d865530f

PR #3826, opened 2024-08-08
Update AndroidImplementation.java

Eric-Chomba:
Delete cached file after sharing it to prevent appending IMG_DATE_TIME to the file name when re-sharing it.

https://www.reddit.com/r/cn1/comments/1ed9ngf/share_api_appending_img_date_time_to_file_name/


shai-almog:
@Eric-Chomba The reformatting to the file makes the change practically unreadable. I get the point of consistency but doing something like this would mean all prior changes would become useless.

The change itself seems fine. Can you commit a version with only that change?

Eric-Chomba:
Okay, create a new pull request

https://github.com/codenameone/CodenameOne/pull/3827

PR #3844, opened 2024-09-29
feat: [3807] Tidy pom.xml files created by generate-app-project goal

### LINKED ISSUE #3807: Format pom.xml of the Generated Maven project
A Generated Maven project at [https://start.codenameone.com/](https://start.codenameone.com/) has unnecessary line breaks in pom.xml of the project's root directory MyApp/myapp/pom.xml which requires a lot of scrolling when studying the pom.xml. Can the unnecessary line breaks be removed as in common/pom.xml?

cn1.plugin.version in MyApp/myapp/pom.xml is generated as 7.0.86. can it be latest CN1 version number by default also?

shannah:
Fixes #3807


shannah:
Closed because found that this issue is caused by a bug in maven that was fixed in 3.6.3.  Updated server to use newer maven resolved the issue.

PR #3855, opened 2024-11-26
Add fireTabEvent

ImmediandoSrl:
I would like to add fireTabEvent to manage Tab key with physical keyboard


shai-almog:
This is a bit problematic. You're changing the global portable API to support a feature that's only supported on very niche android implementations. Tabbing is common, e.g. with the Virtual Keyboard we tab between text fields on Android. Yet this won't fire. It needs to do so on all ports/modes.

The "right thing" would be to listen to focus change events which is the portable API here.

PR #3878, opened 2025-03-06
fix: [3851] fix regression crash in ios text fields

shannah:


shannah:
Need to do some more testing before merging

shannah:
Tested on device with both a text field and text area.  Appears to be working fine now.

DurankGts:
this fix will be available inmediatly or tomorrow?

DurankGts:
> this fix will be available inmediatly or tomorrow?

hello any response about this???

Issue #3904, opened 2025-06-12
new Android UI prevents bottom textareas from going higher when selected


javieranton-zz:
See video

Built with SDK 36 and tested with Samsung S21 (Android 15 and OneUI 7.0)

`setFormBottomPaddingEditingMode(true)` is in use
The new UI introduced Notch checking for devices with top center cameras which messes up the toolbar in the same way iOS notch checking (safe area) did 5 years ago (https://www.codenameone.com/blog/safe-areas.html), but that isn't a problem for me as I am able to work around it. Textareas not following the keyboard is a big problem as that is native built-in functionality

https://github.com/user-attachments/assets/6840e401-b194-4302-96a5-9c484393ea41

The following is a minimally reproducible code:

        
       Form hi = new Form("Test", new BorderLayout());
        Container center = new Container(BoxLayout.y());
        center.setFocusable(false);
        hi.add(BorderLayout.CENTER, center);
        hi.setFormBottomPaddingEditingMode(true);
        TextField msgTxt = new TextField("", "hint", 80, TextField.ANY);
        southContainer = new Container(new BorderLayout());
        msgTxt.setSingleLineTextArea(false);
        msgTxt.setRows(4);
        Button sendBtn = new Button("Send");
        southContainer.add(BorderLayout.CENTER, msgTxt);
        southContainer.add(BorderLayout.EAST, sendBtn);
        hi.add(BorderLayout.SOUTH, southContainer);
        hi.show();

shannah:
Confirmed this affected android 34+.  It was not a regression by CN1, and it did not relate to safe area support.  It was caused by a change in Android behaviour, relating to layout policies in API 34.

shannah:
Fixed.  Will be part of update next Friday.

javieranton-zz:
Thanks for the fast response. I guess my Android OS didn't use API 34+ until I got OneUI 7.0 and that's why I hadn't noticed before, since I have built with API 36 for a few weeks now

DurankGts:
> See video
> 
> Built with SDK 36 and tested with Samsung S21 (Android 15 and OneUI 7.0)
> 
[...]

PR #3905, opened 2025-06-14
fix: [3904] Fixed android async editing with form resizing on API 34+

shannah:
Android 34+ no longer supports the WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE parameter for resizing the form for the soft editor. This broke our in place editor when using formbottompadding editing mode as described in https://github.com/codenameone/CodenameOne/issues/3904.

This fix uses a different approach to add the insets manually.

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java
+++ b/Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java
@@ -58,6 +58,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewParent;
 import android.view.ViewTreeObserver;
+import android.view.Window;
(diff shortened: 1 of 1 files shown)


DurankGts:
hi. I have check that my app today don't resize the Form when the keyboard is open in android 15 an I tested the this the cause. 
image

Before that you submitt this committ https://github.com/codenameone/CodenameOne/pull/3905/commits/f424b059cc0fc1bb863a22e7180c5b22663ecd68 my app it work rezising the Form perfectly but I notice today that don't rezise.

when you submitt some new update you must check if this don't impact other developer because this cause late delivering to our customers in things that was tested

DurankGts:
please @shannah or @shai-almog check this issue

Issue #3909, opened 2025-07-05
Regression Android Async Editing


shannah:
**Describe the bug**
After this fix https://github.com/codenameone/CodenameOne/commit/a87ef18386563d6726035da904f605936c9d8d75#commitcomment-161497322

There is a problem on API 34 with android async editing when immersive mode is disabled.

**To Reproduce**
Haven't been able to reproduce it but has been reported here https://github.com/codenameone/CodenameOne/commit/a87ef18386563d6726035da904f605936c9d8d75#commitcomment-161497322

**Expected behavior**
Keyboard should appear correctly

DurankGts:
this commit don't yet fix the issue in Api34+. The Keyboard hide my TextField or TextArea where there is Container that don't have centerF.setScrollableY(false);

Ajust PAN not always is necesary . For Example in an Form like a chat it's necessary to use Ajust Resize.

DurankGts:
As I explain you  here https://github.com/codenameone/CodenameOne/commit/a87ef18386563d6726035da904f605936c9d8d75#commitcomment-161483180

this method don't fire.

DurankGts:
this a real solutions https://github.com/codenameone/CodenameOne/pull/3911

DurankGts:
this a Test case 

        Form myForm = new Form("Fixed issue 3909 androd resize 34+", new BorderLayout());
        myForm.setFormBottomPaddingEditingMode(false);
        myForm.setScrollableY(false);

        final Container cntList = new Container(BoxLayout.y());
        cntList.setScrollableY(true);

        Container southContainer = new Container(new BorderLayout());
        final TextField msgTxt = new TextField("", "hint", 80, TextField.ANY);
        msgTxt.setSingleLineTextArea(false);

        final Button sendBtn = new Button("Send");
        sendBtn.addActionListener(new ActionListener() {
            @Override
            public void actionPerformed(ActionEvent evt) {
                cntList.add(new Label(msgTxt.getText()));
                cntList.revalidate();
                msgTxt.setText("");
                msgTxt.stopEditing();
            }
        });

        southContainer.add(BorderLayout.CENTER, msgTxt);
        southContainer.add(BorderLayout.EAST, sendBtn);

        myForm.add(BorderLayout.CENTER, cntList);
        myForm.add(BorderLayout.SOUTH, southContainer);
        myForm.show();

expected result in android 14,15

[...]

PR #3911, opened 2025-07-10
100725_1110_AM real solution resize android api34 plus

DurankGts:
this fix this issue https://github.com/codenameone/CodenameOne/issues/3909 and this https://github.com/codenameone/CodenameOne/issues/3904

tested on :
Samsung S24 android 15
Honor RAM turbo android 14
Moto G51 android 12

Checking before create this commit in android 14 and 15 using target 34 in android i was work perfectly. My apps in production don't had this problem of resizing but this this commit https://github.com/codenameone/CodenameOne/pull/3911 solved the issues

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
+++ b/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
@@ -9025,15 +9025,76 @@ public class AndroidImplementation extends CodenameOneImplementation implements 
         if (getActivity() == null) {
             return;
         }
-        if(getCurrentForm() == null){
+        if (getCurrentForm() == null) {
             flushGraphics();
         }
-        if(editInProgress()) {
+        if (editInProgress()) {
             stopEditing(true);
         }
+        resetInsetsForNewForm();// Clean paddings y listeners before Form
         super.setCurrentForm(f);
-        if (isNativeTitle() &&  !(f instanceof Dialog)) {
+        if (isNativeTitle() && !(f instanceof Dialog)) {
             getActivity().runOnUiThread(new SetCurrentFormImpl(getActivity(), f));
+        }
+    }
+    
+    private void resetInsetsForNewForm() {
+        final String methodName = "resetInsetsForNewForm";
+        if (Build.VERSION.SDK_INT >= 34) {
+            final Activity activity = getActivity();
+            if (activity == null) {
+                Log.e(methodName, "Activity is null");
+                return;
+            }
+            activity.runOnUiThread(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        Window window = activity.getWindow();
+                        if (window == null) {
+                            Log.e(methodName, "Window is null");
+                            return;
+                        }
+                        View decorView = window.getDecorView();
+                        ViewGroup content = window.findViewById(android.R.id.content);
+                        if (content == null || content.getChildCount() == 0) {
+                            Log.e(methodName, "Content view is empty");
+                            return;
+                        }
+                        View rootView = content.getChildAt(0);
+                        if (rootView == null || rootView.getWindowToken() == null || !rootView.isAttachedToWindow()) {
+                            Log.e(methodName, "RootView is not ready (null or not attached)");
+                            return;
+                        }
+                        // Restore config by Default
+                        window.setDecorFitsSystemWindows(true);
+                        // Force Refresh
+                        window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_PAN);
+                        window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE);
+                        // Clean listeners and insets
+                        decorView.setOnApplyWindowInsetsListener(null);
+                        decorView.setPadding(0, 0, 0, 0);
+                        decorView.setFitsSystemWindows(true);
+                        decorView.requestApplyInsets();
+
+                        rootView.setOnApplyWindowInsetsListener(null);
+                        rootView.setPadding(0, 0, 0, 0);
+                        rootView.setFitsSystemWindows(true);
+                        rootView.requestApplyInsets();
+
+                        content.requestLayout();
+                        content.invalidate();
+
+                        decorView.requestLayout();
+                        decorView.invalidate();
+
+                        Log.i(methodName, "Insets cleaned successfull");
+
+                    } catch (Exception e) {
+                        Log.e(methodName, "Error appliying insets: " + e.toString(), e);
+                    }
+                }
+            });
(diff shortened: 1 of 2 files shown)


Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
```diff
@@ -9025,17 +9025,78 @@ public void setCurrentForm(final Form f) {
         if (getActivity() == null) {
             return;
         }
-        if(getCurrentForm() == null){
+        if (getCurrentForm() == null) {
```
shai-almog, line 9028: Please try to avoid reformat changes such as these.

Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
```diff
@@ -9025,17 +9025,78 @@ public void setCurrentForm(final Form f) {
...
+                            Log.e(methodName, "Content view is empty");
+                            return;
+                        }
+                        View rootView = content.getChildAt(0);
+                        if (rootView == null || rootView.getWindowToken() == null || !rootView.isAttachedToWindow()) {
+                            Log.e(methodName, "RootView is not ready (null or not attached)");
+                            return;
+                        }
+                        // Restore config by Default
+                        window.setDecorFitsSystemWindows(true);
```
shai-almog, line 9070: This code seems very broad. My concern is that it might fail in some scenarios e.g. when different build hints are used.

What do you think @shannah

shai-almog (changes requested):

Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
DurankGts: I edit your code directly in git hub to avoid your last conversation about my Pull request. I sincerily can't understand why when I try to submit and Improvement a simple diffs cause this conversation..

DurankGts:
this solutions just impact to api 34+ , this means that your source code keep same to others api. I tested using your generated source code to my app using android studio.

Ports/Android/src/com/codename1/impl/android/InPlaceEditView.java
```diff
@@ -1377,26 +1377,28 @@ public void run() {
     }
 
     private static void setEditMode(final boolean resize) {
-        resizeMode = resize;
-
+        resizeMode = true;
```
shai-almog, line 1380: Won't this break `setEditMode(false)`?

It looks like changes to this method would significantly impact backward compatibility.

shai-almog:
As this PR stands it seems problematic. Maybe @shannah has an idea on how to mitigate the risks here.

shannah:
This issue should be fixed with https://github.com/codenameone/CodenameOne/pull/3915

Key points learned:
- API 34 behaves like API 33 and earlier with respect to "immersive mode" unless explicitly opted-into edge-to-edge painting, which we don't, I've adjusted targeting of edge to edge painting workarounds to API 35+.
- Bottom navigation bar on API 35+ is painted over top of the canvas, which is different than <35.  To accommodate this, I have incorporated the bottom navigation bar into safe areas.  So if you have components or containers that should not be painted over by the bottom navigation bar, then they should have `setSafeArea(true)`

DurankGts:
this fix is ready in your built server?

shai-almog:
No. 
[This](https://github.com/codenameone/CodenameOne/pull/3915) was used instead.

DurankGts:
Hi. Again...

with this fix #3915  check this different behaivor between android version an ios

image


this my test case

    public void textAreaIssueInAndroid() {
        Form myForm = new Form("Fixed issue 3909 androd resize 34+", new BorderLayout());
        myForm.setFormBottomPaddingEditingMode(false);
        myForm.setScrollableY(false);
        **myForm.getToolbar().setSafeArea(true); //Impact  in android 15 side if the toolbar Uiid has padding** this line always has been by default in all my forms, but now if I want to set before the new fix #3915 a problem is generate in ios.


        final Container cntList = new Container(BoxLayout.y());
        cntList.setScrollableY(true);

        Container southContainer = new Container(new BorderLayout());
        //southContainer.setSafeArea(false); //if I uncomment the bottom spaces is fixed in ios but in android 15 other gap or padding is added to botom.
        final TextField msgTxt = new TextField("", "hint", 80, TextField.ANY);
        msgTxt.setSingleLineTextArea(false);

        final Button sendBtn = new Button("Send");
        sendBtn.addActionListener(new ActionListener() {
            @Override
                cntList.add(new Label(msgTxt.getText()));
                cntList.revalidate();
                msgTxt.setText("");
                msgTxt.stopEditing();
            }
        });

        southContainer.add(BorderLayout.CENTER, msgTxt);
        southContainer.add(BorderLayout.EAST, sendBtn);

        myForm.add(BorderLayout.CENTER, cntList);
        myForm.add(BorderLayout.SOUTH, southContainer);
        myForm.show();
    }


any solutions to this different now?. Before this solutions #3915  all work perfectly in android 15 and others. I think that in android 15 your are calulating the same gap of the system tool bar up and the same gap os the bottom bar althought the screen isn't in full screen that is the issue with inmersive by default in android 15. see the picture.

shannah:
You need

```
southContainer.setSafeArea(true);
```

That will add bottom padding on iOS.

You probably shouldn't have 

```
myForm.getToolbar().setSafeArea(true);
```

What device is the Android 15 in the screenshot?

DurankGts:
> You need
> 
> ```
> southContainer.setSafeArea(true);
> ```
> 
> That will add bottom padding on iOS.
> 
> You probably shouldn't have
> 
> ```
> myForm.getToolbar().setSafeArea(true);
> ```
> 
> What device is the Android 15 in the screenshot?

image

DurankGts:
this the result in android 15 when  I use southContainer.setSafeArea(true);

image

shannah:
This is what I get on Android 15.

(But without the `myForm.getToolbar().setSafeArea(true);`)

image

What device are you using?

DurankGts:
this the behaivor of your resizing please check the video https://youtube.com/shorts/aBqWXJJMwbE

DurankGts:
Any response about this issue @shannah ?

DurankGts:
I was checking this commit https://github.com/codenameone/CodenameOne/pull/3915/commits/65c6e19ee8087f08788028c9839dd31e539675f6 I suppose that this will fix my issue in android 15 sure? I have checked that you just check if the Keyboard is Open to recalculate the SafeArea. My question is if you don't check if the App **is not in  **Full Screen**** like this case to avoid unnecessary Gap Up and Bottom althought the keyboard isn't open like this case https://github.com/codenameone/CodenameOne/pull/3911#issuecomment-3070835117?

Issue #3888, opened 2025-03-24
Android bug: Title/Toolbar from Form is too high on screen


schreiby:
**Describe the bug**
Title and toolbar in the form is too high on the screen so user can't click any toolbar buttons

**To Reproduce**
1. Run this on Pixel 9 emulator and you see the title this way even if you set SafeArea to true

**Expected behavior**
I hope that the title and toolbar would not be so high on the screen so the camera doesn't "cut off" the words and the toolbar buttons/commands will actually be clickable.

**Screenshots**
![Image](https://github.com/user-attachments/assets/4d4b4a90-b186-4ca3-860a-f8304c552701)

**Smartphone (please complete the following information):**
 - Device: Pixel 9 (real device + simulator)
 - OS: Android 15

**Additional context**
Please advise me since I tried setting the SafeArea and SafeArea root to true and this did not seem to help. Toolbar commands/buttons are not responding to user when clicked. It also seems that the toolbar/title are overlapping with the device's Statusbar. 

Thank you.

schreiby:
Thank you for assigning this. I'm really grateful. 

Just wondering if there's any quick workaround I can use till there's a real fix? 

Thanks.

shannah:
Fixed.  https://github.com/codenameone/CodenameOne/pull/3889

Will be part of release on Friday.

schreiby:
Thank you!

DurankGts:
about this issue you must to implement a way to avoid impact to others Developer when you implements issue la UI. Now my app have extra space in all may forms. Doy you implement a build hint to avoid this???

DurankGts:
now alls may Forms in android have this unnecesary space

![Image](https://github.com/user-attachments/assets/0d4820dd-a9f6-400a-b5a9-ede32a45a2fd)

shannah:
Can you share your build hints?

I'm not reproducing this.

Pixel 9

![Image](https://github.com/user-attachments/assets/093403bb-e406-470c-b34f-adc4e4f7b407)

Pixel 2

![Image](https://github.com/user-attachments/assets/6558883d-34b0-46fb-b33b-72c7f6058d7e)

Pixel

![Image](https://github.com/user-attachments/assets/ee92a852-b1ca-427f-856b-d9808edb47f4)

DurankGts:
this my build hints

#
#Tue Apr 01 18:30:19 VET 2025
codename1.arg.android.debug=false
[...]

Issue #3916, opened 2025-07-13
New Android south textbox is still too high


javieranton-zz:
7.0.198 didn't fix the following issue: #3913

I am now using both 

`super.getAllStyles().setPaddingBottom(getPlatformName().equals("and") ? 145 : 40);`

And

`CN.setProperty("android.useSafeAreaInsets", "true");`

But the textbox is still too high

If I drop the bottom padding, I am back to square one having the form extend down to the system navigation buttons

shannah:
You should need neither of those

shannah:
As far as i can tell it should be working correctly now.  please try the test case i provided.  If that doesn't reproduce the issue, please modify it so that the issue is reproducible.

shannah:
This is not what i get on this test case. 

The explicit padding shouldn't be needed.

The result you're getting looks like the result i can reproduce with the previous release,

Can you do a build with sources then find the UpdateSafeArea() method on CodenameOneView.java and post it here?  I want to verify that it is using the latest.

javieranton-zz:
```
public static boolean isImmersive() {
        if (getActivity() == null) {
            return false;
        }
        return isImmersive(getActivity().getWindow());
    }
    public static boolean isImmersive(Window window) {
        if (Build.VERSION.SDK_INT >= 35) {
            // Android 15+ is always immersive (overlay mode by default)
            return true;
        }
        // On Android 34 and below, we can't detect decorFitsSystemWindows
        // reliably at runtime. So the app must make the decision explicitly.
        return false;
    }
```

Built with SDK 36 and tested with Samsung S21 (Android 15 and OneUI 7.0)

shannah:
Hmm.  That looks right.  I've been relying on emulators, but that test case was working fine with the api 30, 33, 34, 35, and 36 virtual devices.

I'll try to reproduce again.  Can you post your test case so i can be sure we're using the same?

javieranton-zz:
Sure. Here is the test case (same as yours but with padding added)

```
Form myForm = new Form("Fixed issue 3909 androd resize 34+", new BorderLayout());
        myForm.setFormBottomPaddingEditingMode(false);
        myForm.setScrollableY(false);
        myForm.getAllStyles().setPaddingBottom(getPlatformName().equals("and") ? 145 : 40);
        final Container cntList = new Container(BoxLayout.y());
        cntList.setScrollableY(true);

        Container southContainer = new Container(new BorderLayout());
[...]

Issue #3914, opened 2025-07-12
Race condition when using android.useSafeAreaInsets property


shannah:
On launch, sometimes extra padding appears at the top of the form when using android.useSafeAreaInsets

<img width="484" height="603" alt="Image" src="https://github.com/user-attachments/assets/1907374d-97f8-4f09-a933-c309809c7591" />

This seems to be a race condition since the safe area insets need to be calculated on the main thread, and it doesn't revalidate the canvas after setting them.

shannah:
https://github.com/codenameone/CodenameOne/pull/3915 should fix this issue.  It issues a revalidate if the safe area is changed.

PR #3918, opened 2025-07-15
150725_1114_AM real fix gap safe area android side 15 and others versions

DurankGts:
🐞 **Bug Summary**
In Android 11+ (including Android 15), when setDecorFitsSystemWindows(true) is active (the default), the system already applies padding for system bars (status bar, navigation bar, gesture area).

The original updateSafeArea() logic manually adds those same systemBarInsets again, resulting in double padding on the top and bottom of the screen — even when the keyboard is not visible.

This caused unwanted blank space above and below the Form content on Android 13–15.

✅ **Fix Summary**
The fix improves AndroidImplementation.isImmersive(Window) to check the actual runtime state of decorFitsSystemWindows.
Then, in updateSafeArea():

Insets are only added when Android is not already applying them (i.e. in immersive/overlay mode).

If the system already applied them, the manual padding is subtracted.

This prevents double insets and ensures consistent layout across Android versions.

✅ **Why this improvement is necessary and compatible with Android 15 and below**

📌 **Problem**
In Android 11 (API 30) and above, the method Window.setDecorFitsSystemWindows(false) allows apps to take full control of layout insets. **When this flag is set to false, Android no longer automatically applies padding for system bars (status bar, navigation bar, gesture area, etc.).**

Codename One's original implementation of isImmersive() assumed that:

Immersive mode (i.e., manual inset handling) is only enabled on Android 15+, and

Lower versions do not support this and thus always apply insets automatically.

**This logic fails in cases where:**

The developer sets setDecorFitsSystemWindows(false) manually on Android 11–14.

The framework applies system insets and the app adds them again, **causing duplicated margins or padding.** like this https://github.com/codenameone/CodenameOne/pull/3911#issuecomment-3070835117 and this **https://github.com/codenameone/CodenameOne/issues/3916#issuecomment-3067296508**

✅ **What this improvement does**
The new implementation of AndroidImplementation.isImmersive(Window window):

Uses reflection to check window.getDecorFitsSystemWindows() at runtime (API 30+).

If getDecorFitsSystemWindows() returns false, it means manual inset compensation is required.

If reflection is not available or fails (e.g., on older devices), the app can explicitly override immersive mode using a system property:

**Display.getInstance().setProperty("android.forceImmersive", "true");**
This logic is now used inside CodenameOneView.updateSafeArea() to **avoid double-applying system insets.**

🧩 **Compatibility**
✅ Android 15+ (API 35+): Still works as before, since these versions default to overlay behavior.

✅ Android 11–14 (API 30–34): Now correctly respects the developer’s use of setDecorFitsSystemWindows(false).

✅ Android 10 and below (API < 30): Safe fallback using the property override.

✅ No changes needed in user code unless they explicitly want to force immersive mode on old versions.

✅ **Result**
This change ensures that the safe area insets are:

Only applied **when needed** (i.e., when Android isn’t doing it already).

Prevents layout **bugs caused by double-padding.**

**Keeps Codename One apps visually consistent across Android versions.**

### DIFF
--- a/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
+++ b/Ports/Android/src/com/codename1/impl/android/AndroidImplementation.java
@@ -1299,13 +1299,19 @@ public class AndroidImplementation extends CodenameOneImplementation implements 
         return isImmersive(getActivity().getWindow());
     }
     public static boolean isImmersive(Window window) {
-        if (Build.VERSION.SDK_INT >= 35) {
-            // Android 15+ is always immersive (overlay mode by default)
-            return true;
+        if (Build.VERSION.SDK_INT >= 30) {
+            try {
+                // Android 11+ supports decorFitsSystemWindows
+                Method m = Window.class.getMethod("getDecorFitsSystemWindows");
+                Boolean result = (Boolean) m.invoke(window);
+                return result != null && !result;  // Immersive = decorFitsSystemWindows == false
+            } catch (Throwable t) {
+                // Fallback: allow developer override
+                return "true".equals(Display.getInstance().getProperty("android.forceImmersive", "false"));
+            }
         }
-        // On Android 34 and below, we can't detect decorFitsSystemWindows
-        // reliably at runtime. So the app must make the decision explicitly.
-        return false;
+        // For Android < 11, use only developer override
+        return "true".equals(Display.getInstance().getProperty("android.forceImmersive", "false"));
     }
     public static Rect getSystemBarInsets(final View rootView) {
         final Rect result = new Rect(0, 0, 0, 0);
--- a/Ports/Android/src/com/codename1/impl/android/CodenameOneView.java
+++ b/Ports/Android/src/com/codename1/impl/android/CodenameOneView.java
@@ -164,9 +164,9 @@ public class CodenameOneView {
 
     private void updateSafeArea() {
         final Activity activity = CodenameOneView.this.implementation.getActivity();
-
         final Rect rect = this.safeArea;
         final View rootView = activity.getWindow().getDecorView();
+
         if (Build.VERSION.SDK_INT >= VERSION_CODE_P) {
             try {
                 Method getRootWindowInsetsMethod = View.class.getMethod("getRootWindowInsets");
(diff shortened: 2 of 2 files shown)


shai-almog:
We know how to use an LLM too. They are great at writing a wall of text which we then need to read and review and discover that they generated code that doesn't work and even references methods that don't exist. This means that instead of saving work we wasted time reviewing bad PRs generated by an LLM instead of working on fixing the actual problem.

DurankGts:
> We know how to use an LLM too. They are great at writing a wall of text which we then need to read and review and discover that they generated code that doesn't work and even references methods that don't exist. This means that instead of saving work we wasted time reviewing bad PRs generated by an LLM instead of working on fixing the actual problem.

I tested all my PR in real devices before submit a solution to you. Why you Just use a draft branch a test if my PR it work before close my branch? I can't understand why you simply rejected a real solution that late or early you will need and others developers as my case that are waiting maybe 15 days to submit a new update to Google market, but this is blocking me to deliver a release. 

I will never submit a solutions that doesn't work...

DurankGts:
@shannah or @shai-almog  can you provide a solutions of gap up/bottom on android 15 to the closed issue #3888 , #3916 ,#3914 that is impacting my entire app now?

Like you rejected this PR https://github.com/codenameone/CodenameOne/pull/3918 that was tested on real devices and fix this solution. Some times I thinks that I'm the unic developer testing on android15? I would like to Know when I will submit my release to Google Markets to our customers. I need to submit 9 Apps before the 31 of august to necessary increase to api 35 from google and this issue resizing and gap (up/bottom) is blocking the user experience.

DurankGts:
I think that I found the issue. I was using **codename1.arg.android.targetSDKVersion=34** and I was testing on real device Api35 and this create a gap up/bottom in all my app. 

When I increase the codename1.arg.android.targetSDKVersion=35 **then the gap dissapear an all is ok.** The other issue now is that in android 15 the toolbar color keed light in entire screen and when my Form is White complete the the system bar don't show. This other improvement from API 35. **Theres a way to keep the system toolbar in dark background and light text?**

DurankGts:
the pr #3918 fix all this issue but if you don't want don't use. All the developers must be to update to api 34 to avoid this issue.

Thank you and excuse me my comments

DurankGts:
the unic problem increasing to Api 35 where is necessary use edge to edge screen is that the developer need to manage the System toolbar color in your forms. In my case i Solve using my costom native interface to update the system toolbar color.

PR #3981, opened 2025-10-17
Checkstyle fixes second round

shai-almog:


shai-almog:

PR #4073, opened 2025-10-24
Enforce console output checks in core unit tests

shai-almog:
## Summary
- add an auto-registered JUnit 5 extension that captures console output during tests and fails when unexpected messages appear
- introduce an `@AllowConsoleOutput` annotation so specific tests can opt in to logging when required
- register the extension via service loader configuration so it applies to every core unit test run

## Testing
- mvn test *(fails: missing dependency com.codenameone:codenameone-factory:jar:8.0-SNAPSHOT)*

------
https://chatgpt.com/codex/tasks/task_e_68fb98e861f08331bc10960d53dfc050


shai-almog:

Issue #4000, opened 2025-10-20
Local Build Release Apk With CircleProgress Doesn't Open


Eric-Chomba:
Using the following test case, I created a form which shows CircleProgress animation

```

    public void runApp() {
        Form hi = new Form("Hi World", BoxLayout.y());
        Button helloButton = new Button("Hello World");
        hi.add(helloButton);
        helloButton.addActionListener(e -> hello());
        hi.getToolbar().addMaterialCommandToSideMenu("Hello Command",
                FontImage.MATERIAL_CHECK, 4, e -> hello());
        hi.add(cntAnimation("Loading..."));
        hi.show();

    }

    public Container cntAnimation(String message) {
        Container cntAnimation = new Container(BoxLayout.y());
        SpanLabel lblName = new SpanLabel("placeholder", "CenterAlignmentLabel");
        final int circleSize = Display.getInstance().convertToPixels(5);
        Container cntName = new Container(new BorderLayout()) {
            @Override
            protected Dimension calcPreferredSize() {
                return new Dimension(circleSize, circleSize);
            }
        };

        cntName.add(BorderLayout.CENTER, lblName);
        Container cntDemo = BorderLayout.centerCenter(cntName);
        //replace the label with a CircleProgress to indicate that it is loading
        CommonProgressAnimations.CircleProgress.markComponentLoading(lblName).setUIID("circleAnim");
        cntAnimation.add(FlowLayout.encloseCenterMiddle(new SpanLabel(message, "lblAnim")));
        cntAnimation.add(cntDemo);
        return cntAnimation;
    }
```

This works in server debug & release apk builds. Using a locally generated gradle project, it also works in a local debug apk build but fails for local release apk build. 

It shows the following exception in Android studio logcat.

```
java.lang.RuntimeException: Failed to create progress component: java.lang.Class<r0.h> cannot be instantiated
at com.codename1.ui.CommonProgressAnimations$ProgressAnimation.I5(CommonProgressAnimations.java:47)
at com.codename1.ui.CommonProgressAnimations$CircleProgress.J5(CommonProgressAnimations.java:3)
at com.example.myapp.MyApp.l(MyApp.java:47)
at com.example.myapp.MyApp.g(MyApp.java:51)
at com.codename1.system.Lifecycle.h(Lifecycle.java:9)
at com.example.myapp.MyAppStub.run(MyAppStub.java:43)
at com.example.myapp.MyAppStub$1.run(MyAppStub.java:17)
[...]

PR #4081, opened 2025-10-26
Disable R8 full mode in android gradle projects

shannah:
This was causing issues like https://github.com/codenameone/CodenameOne/issues/4000


shai-almog:

PR #4445, opened 2026-01-19
Resolve invokespecial owners for interface default methods

shai-almog:
### Motivation
- Fix incorrect `invokespecial` target selection that caused generated C calls to reference the wrong declaring class for inherited default/interface methods (e.g. `super.drawNumber` resolving to the wrong parent). 

### Description
- Add `ByteCodeClass.findMethodOwner(name, desc)` to search the class, its superclass chain and interfaces for the first non-abstract declaring owner and avoid cycles via a visited set. 
- Add `Util.resolveInvokeSpecialOwner(owner, name, desc)` helper to map the original owner to the resolved declaring class name. 
- Use the new helper in both normal and custom invocation code paths by remapping `owner` when emitting `INVOKESPECIAL` (excluding `<init>` and `<clinit>`), so super calls target the real concrete implementation. 
- Add a small `findDeclaredMethod` helper to locate a declared method on a class.

### Testing
- No automated tests were run for this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_696e3befc30c8331acb04dee99a6ade8)

### DIFF
--- a/vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeClass.java
+++ b/vm/ByteCodeTranslator/src/com/codename1/tools/translator/ByteCodeClass.java
@@ -289,6 +289,44 @@ public class ByteCodeClass {
         return false;
     }
 
+    public ByteCodeClass findMethodOwner(String name, String desc) {
+        return findMethodOwner(name, desc, new HashSet<ByteCodeClass>());
+    }
+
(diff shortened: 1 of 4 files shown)


ddyer0:
Here's another failed build program, based on a different hierarchy, but with similar features of using and overriding
default methods.  I had to add a lot of methods to the naked skeleton of the classes to get it to fail.  If this still fails
for you, I'm sure you'll have fun figuring it out :)
[Dtest.java](https://github.com/user-attachments/files/24723920/Dtest.java)

PR #4606, opened 2026-03-11
Prevent NPE when refreshing active geofences

shai-almog:
### Motivation
- Fix a regression where `GeofenceManager.update(..., true)` could hit a `NullPointerException` when the in-memory `activeFences` map was not initialized but persisted state was missing.

### Description
- Initialize the cached `activeFences` map early in `GeofenceManager.update(boolean, Location)` by calling `getActiveFences(false)` before any removals to avoid null dereference.
- Add a regression test `updateForceRefreshHandlesMissingPersistedActiveFenceMap` in `GeofenceManagerTest` that deletes the persisted active-fence storage key and verifies a forced update does not throw and removes the geofence.

### Testing
- Ran the unit test module for the `GeofenceManager` test with Java 8 using `mvn -pl core-unittests -am -DfailIfNoTests=false -DunitTests=true -Dmaven.javadoc.skip=true -Dtest=GeofenceManagerTest test` and the `GeofenceManagerTest` suite passed (all tests succeeded).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b1936d16b8833199e35f3f34a0bc5e)

### DIFF
--- a/CodenameOne/src/com/codename1/location/GeofenceManager.java
+++ b/CodenameOne/src/com/codename1/location/GeofenceManager.java
@@ -502,6 +502,7 @@ public final class GeofenceManager implements Iterable<Geofence> {
         LocationManager.getLocationManager().setBackgroundLocationListener(null);
         List<String> activeIds = new ArrayList<String>(getActiveKeys(false));
         List<String> activeKeys = getActiveKeys(false);
+        Map<String, Geofence> activeFences = getActiveFences(false);
         for (String id : activeIds) {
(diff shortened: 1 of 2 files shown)


shai-almog:
Compared 32 screenshots: 32 matched.

### Benchmark Results

- **Compilation Time:** 259 seconds

| Metric | Duration |
| --- | --- |
| Simulator Boot | 74000 ms |
| App Install | 15000 ms |
| App Launch | 7000 ms |
| Test Execution | 151000 ms |

PR #4750, opened 2026-04-14
Fix concrete invoke dependency resolution when method context is null

liannacasper:
### Motivation
- Dependency collection could run without a method context, and `resolveConcreteInvokeOwner()` previously bail/returned `null` when `getMethod()` was unavailable which caused concrete-class header/includes to be skipped and later produced calls to targets without declarations.

### Description
- Stop requiring `getMethod()` to be non-null in `Invoke.resolveConcreteInvokeOwner()` and `CustomInvoke.resolveConcreteInvokeOwner()` so concrete-owner resolution can run during dependency analysis even when no method context exists.
- Make the `currentClass` lookup null-safe and keep the self/inner-class guard only when `currentClass` is present.
- Changes touch `vm/ByteCodeTranslator/src/com/codename1/tools/translator/bytecodes/Invoke.java` and `.../CustomInvoke.java` with small, equivalent edits in both files.

### Testing
- Ran `mvn -f vm/ByteCodeTranslator/pom.xml -DskipTests compile` which succeeded. 
- Attempted the Java 8 setup `export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64 && export PATH="$JAVA_HOME/bin:$PATH" && mvn -f vm/ByteCodeTranslator/pom.xml -DskipTests compile` but that failed because the documented Java 8 path is not available in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69de462d9088832997b65291cf18b878)

### DIFF
--- a/vm/ByteCodeTranslator/src/com/codename1/tools/translator/bytecodes/CustomInvoke.java
+++ b/vm/ByteCodeTranslator/src/com/codename1/tools/translator/bytecodes/CustomInvoke.java
@@ -94,7 +94,7 @@ public class CustomInvoke extends Instruction {
         String dependencyOwner = owner;
         if (origOpcode == Opcodes.INVOKEVIRTUAL) {
             ByteCodeClass bc = Parser.getClassObject(owner.replace('/', '_').replace('$', '_'));
-            String resolvedConcreteOwner = resolveConcreteInvokeOwner(bc);
+            String resolvedConcreteOwner = resolveConcreteInvokeOwner(bc, true);
             if (resolvedConcreteOwner != null) {
                 dependencyOwner = resolvedConcreteOwner;
             }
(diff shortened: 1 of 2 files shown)


shai-almog:
Compared 37 screenshots: 37 matched.

### Benchmark Results

- **Compilation Time:** 114 seconds

| Metric | Duration |
| --- | --- |
| Simulator Boot | 1000 ms |
| Simulator Boot (Run) | 0 ms |
| App Install | 2000 ms |
| App Launch | 3000 ms |
| Test Execution | 170000 ms |
| Base64 native encode | 1962.000 ms |
| Base64 CN1 encode | 2123.000 ms |
| Base64 encode ratio (CN1/native) | 1.082x (8.2% slower) |
| Base64 native decode | 1270.000 ms |
| Base64 CN1 decode | 1316.000 ms |
| Base64 decode ratio (CN1/native) | 1.036x (3.6% slower) |

PR #4769, opened 2026-04-17
iOS: Defer updateCanvas on foreground to avoid iPad display size flip

### LINKED ISSUE #4767: Ipad frame size after restart
On ipad, if you push the "home" button and then restart the app, the width (or height in portrait mode)
is reset to an incorrect value.   Curiously, this acts as a toggle - if you do it twice it gets reset to the correct
value.

The attached screen shot has a bunch of numbers recording the flow of events.  Note that the
reset of width to 1536 (instead of 2048) occurs between the "stop" and "restart" events.  

Image

liannacasper:
### Motivation
- On iPad the app could report the short edge as the display width/height when resuming from background, producing a transient incorrect screen-size event (e.g. 1536 vs 2048) between stop/restart lifecycle callbacks.

### Description
- Change in `Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m`: `cn1ApplicationWillEnterForeground` now defers the call to `updateCanvas:YES` using `dispatch_async(dispatch_get_main_queue(), ^{ ... })` so UIKit can settle the root view bounds before the canvas is updated.

### Testing
- Attempted to run the quick smoke script `./scripts/fast-core-unit-smoke.sh` with `JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64`, but the environment does not contain Java 8 so the test run failed and no automated tests were executed in this container.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69e1b84430d883298c741b349e0a08df)

### DIFF
--- a/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
+++ b/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
@@ -218,7 +218,16 @@ static void installSignalHandlers() {
     com_codename1_impl_ios_IOSImplementation_applicationWillEnterForeground__(CN1_THREAD_GET_STATE_PASS_SINGLE_ARG);
     CodenameOne_GLViewController* vc = [CodenameOne_GLViewController instance];
     if (vc != nil) {
-        [vc updateCanvas:YES];
+        // Defer updateCanvas by one run-loop cycle so that UIKit has a chance to
+        // settle the root view bounds after foreground transition.  Calling this
+        // too early can report the short edge for width/height on iPad, which
+        // causes a transient wrong screen-size event between stop/start.
+        dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


shai-almog:
Compared 37 screenshots: 37 matched.

### Benchmark Results

- **Compilation Time:** 114 seconds

| Metric | Duration |
| --- | --- |
| Simulator Boot | 0 ms |
| App Install | 1000 ms |
| App Launch | 6000 ms |
| Test Execution | 184000 ms |
| Base64 native encode | 1908.000 ms |
| Base64 CN1 encode | 1916.000 ms |
| Base64 encode ratio (CN1/native) | 1.004x (0.4% slower) |
| Base64 native decode | 1078.000 ms |
| Base64 CN1 decode | 1363.000 ms |
| Base64 decode ratio (CN1/native) | 1.264x (26.4% slower) |

PR #4837, opened 2026-04-30
CI: share iOS port build across iOS workflows via reusable workflow

liannacasper:
## Summary

The biggest remaining iOS performance gap from the original audit: **the iOS port is built from scratch in every iOS workflow on every PR** (5 builds across `scripts-ios.yml`, `scripts-ios-native.yml`, and `ios-packaging.yml`'s 3-row matrix). This PR makes that build run **at most once per workflow run**, and shares the result across workflows when source hasn't changed.

## How

A new reusable workflow `.github/workflows/_build-ios-port.yml` (workflow_call) runs setup-workspace + build-ios-port. Each of the 3 iOS workflows now has 2 jobs:

```
build-port (uses _build-ios-port.yml)  →  test job (needs: build-port)
```

The test job just restores the populated caches and runs `build-ios-app.sh` + tests. It uses `actions/cache/restore@v4` with `fail-on-cache-miss: true` so a missing cache is a clear error, not silent regression.

## The cache that does the heavy lifting

A new `cn1-built` cache stores the built CN1 + iOS port artifacts:

- `~/.m2/repository/com/codenameone`
- `Themes/`
- `Ports/iOSPort/nativeSources/`

Keyed on a composite hash of:
- All Java / Objective-C / theme source under `CodenameOne/src`, `Ports/iOSPort`, `vm/JavaAPI`, `vm/ByteCodeTranslator`, `Themes/`
- Every `pom.xml`
- `setup-workspace.sh`, `build-ios-port.sh`, `build-native-themes.sh`, and the reusable workflow itself

**No `restore-keys`** — exact match only. A stale artifact set never satisfies a different source state.

The cn1-built cache is restored *after* the broader `~/.m2/repository` cache so it overwrites any stale `com/codenameone/` subtree pulled in by the POM-keyed m2 cache.

On cache hit, `build-port` skips both `setup-workspace` and `build-ios-port` and the job finishes in ~1–2 min instead of ~30 min. The 3 test jobs then restore the cache and run.

## Cross-workflow benefit (the M6 part)

The `cn1-built-${runner.os}-${source-hash}` cache key is identical across all 3 iOS workflows. When scripts-ios.yml runs first on a PR and populates the cache, ios-packaging.yml and scripts-ios-native.yml later in the same PR cache-hit and skip their entire iOS port rebuild.

A PR that doesn't touch CN1 / iOS port source at all will hit the cache from master's last successful run.

## Expected savings

Rough timings:

| Scenario | Before | After |
|---|---|---|
| First run, source changed | ~33 min × 3 matrix in parallel = ~37 min wall, ~99 min CPU | ~30 min build-port → ~10 min × 3 in parallel = ~40 min wall, ~60 min CPU |
| Subsequent run, source unchanged | ~37 min wall, ~99 min CPU | ~2 min build-port → ~10 min × 3 in parallel = ~12 min wall, ~32 min CPU |
| Cross-workflow on same PR | 3 × ~33 min = ~99 min macOS minutes | First workflow: ~33 min, others hit cache: ~12 min each → ~57 min total |

So the warm-cache path (the common case for typical PRs) goes from ~37 min wall-clock to ~12 min, and three iOS workflows on the same PR collectively go from ~3 × 33 min to ~33 + 12 + 12 min.

## Bonus: parparvm-tests.yml self-trigger bug

`parparvm-tests.yml`'s `paths:` filter didn't include itself, so my changes to that workflow in #4836 never actually triggered the workflow's own validation. Fixed by adding `.github/workflows/parparvm-tests.yml` to its `paths:` list.

## Risks / things to watch

- **`fail-on-cache-miss: true` on the test jobs.** If the cn1-built cache somehow doesn't survive the build-port → test-job handoff (e.g., GitHub cache eviction in the seconds between jobs), the test jobs fail loudly. This is intentional — silent fall-through could mask a problem.
- **Hash completeness.** The source hash covers Java, Objective-C, theme, properties, XML, CSS files plus all POMs and the build scripts. If a build output changes based on a file we didn't include, we'd serve a stale artifact. The composite is conservative; happy to tighten or loosen if you spot a gap.
- **First-run wall-clock is slightly slower** (~40 min vs ~37 min) because build-port is serial before the matrix. The CPU savings (~40 min less per first run) offset that, and every subsequent run is dramatically faster.
- **Within-PR cancellation.** The reusable workflow has its own `concurrency:` group (`mac-ios-port-${{ github.workflow }}-${{ github.ref_name }}`) so a new push cancels the in-flight build-port too.

## Test plan

- [ ] All 5 modified/new workflows parse (verified locally with `yaml.safe_load`).
- [ ] First PR run (cold cache): all 3 iOS workflows succeed; `build-port` does the full build, test jobs find the cn1-built cache and skip ahead to `build-ios-app.sh`.
- [ ] Second push on same PR (warm cache): build-port logs "cn1-built cache HIT — setup-workspace and build-ios-port were skipped." Test jobs run as before.
- [ ] Verify cross-workflow share: if scripts-ios.yml runs first then ios-packaging.yml runs later on the same PR, ios-packaging.yml's build-port logs a cache hit.
- [ ] Confirm parparvm-tests.yml runs on this PR (since it modifies `.github/workflows/parparvm-tests.yml`, which is now in the paths filter).

## Out of scope (kept as future work)

- M3/M4: Dedupe the Maven + Ant double-build inside `pr.yml`.
- L6: Drop redundant `push:` triggers on test-only workflows.
- Apply the same reusable-workflow pattern to scripts-android.yml + scripts-javase.yml.

### DIFF
--- a/.github/workflows/_build-ios-port.yml
+++ b/.github/workflows/_build-ios-port.yml
@@ -1,54 +1,28 @@
-name: Test iOS packaging
+name: _Build iOS port (reusable)
+
+# Reusable workflow that runs scripts/setup-workspace.sh + scripts/build-ios-port.sh
+# once per workflow run, populating shared caches that downstream test jobs (in
+# scripts-ios.yml, scripts-ios-native.yml, ios-packaging.yml) restore via the
+# same cache keys.
+#
+# A separate "cn1-built" cache is keyed on the CN1 source hash with no restore-
+# keys (exact match only). On hit, the entire setup-workspace + build-ios-port
+# sequence is skipped — the iOS port artifact in ~/.m2/repository/com/codenameone
+# is already correct for the current source state.
+#
+# This cache is shared across all three iOS workflows on the same branch, so
+# whichever workflow runs first populates it and the others skip the rebuild.
 
 on:
-  pull_request:
-    paths:
-      - '.github/workflows/ios-packaging.yml'
-      - '.github/workflows/_build-ios-port.yml'
-      - 'maven/codenameone-maven-plugin/**'
-      - 'vm/ByteCodeTranslator/**'
-      - 'Ports/iOSPort/**'
-      - 'scripts/build-ios-app.sh'
-      - 'scripts/run-ios-ui-tests.sh'
-      - 'scripts/run-ios-native-tests.sh'
-      - 'scripts/ios/**'
-      - 'scripts/hellocodenameone/**'
-      - '!docs/**'
-  push:
-    branches: [ master ]
-    paths:
-      - '.github/workflows/ios-packaging.yml'
-      - '.github/workflows/_build-ios-port.yml'
-      - 'maven/codenameone-maven-plugin/**'
-      - 'vm/ByteCodeTranslator/**'
-      - 'Ports/iOSPort/**'
-      - 'scripts/build-ios-app.sh'
-      - 'scripts/run-ios-ui-tests.sh'
-      - 'scripts/run-ios-native-tests.sh'
-      - 'scripts/ios/**'
-      - 'scripts/hellocodenameone/**'
-      - '!docs/**'
+  workflow_call:
 
 jobs:
-  build-port:
-    uses: ./.github/workflows/_build-ios-port.yml
-
-  packaging:
-    needs: build-port
-    permissions:
-      contents: read
+  build:
     runs-on: macos-15
-    timeout-minutes: 45
-    # Exercises both CocoaPods and SPM in a single Xcode project. Catches
-    # regressions in either dependency manager (each pathway runs end to end)
-    # and additionally validates that they coexist correctly.
-    env:
-      IOS_DEPENDENCY_ARGS: >-
-        -Dcodename1.arg.ios.dependencyManager=both
-        -Dcodename1.arg.ios.pods=AFNetworking
-        -Dcodename1.arg.ios.spm.packages=swift-collections|https://github.com/apple/swift-collections.git|from:1.1.0
-        -Dcodename1.arg.ios.spm.products.swift-collections=Collections
-
+    timeout-minutes: 60
+    concurrency:
+      group: mac-ios-port-${{ github.workflow }}-${{ github.ref_name }}
+      cancel-in-progress: true
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


shai-almog:
Compared 85 screenshots: 85 matched.

### Benchmark Results

- **Compilation Time:** 181 seconds

#### Build and Run Timing
| Metric | Duration |
| --- | --- |
| Simulator Boot | 64000 ms |
| App Install | 25000 ms |
| App Launch | 7000 ms |
| Test Execution | 287000 ms |

| Metric | Duration |
| --- | --- |
| Base64 native encode | 1540.000 ms |
| Base64 CN1 encode | 1535.000 ms |
| Base64 encode ratio (CN1/native) | 0.997x (0.3% faster) |
| Base64 native decode | 1099.000 ms |
| Base64 CN1 decode | 1673.000 ms |
| Base64 decode ratio (CN1/native) | 1.522x (52.2% slower) |
| Base64 SIMD encode | 578.000 ms |
| Base64 encode ratio (SIMD/native) | 0.375x (62.5% faster) |
| Base64 encode ratio (SIMD/CN1) | 0.377x (62.3% faster) |
| Base64 SIMD decode | 489.000 ms |
| Base64 decode ratio (SIMD/native) | 0.445x (55.5% faster) |
| Base64 decode ratio (SIMD/CN1) | 0.292x (70.8% faster) |
| Image encode benchmark iterations | 100 |
| Image createMask (SIMD off) | 64.000 ms |
| Image createMask (SIMD on) | 15.000 ms |
| Image createMask ratio (SIMD on/off) | 0.234x (76.6% faster) |
| Image applyMask (SIMD off) | 141.000 ms |
| Image applyMask (SIMD on) | 87.000 ms |
| Image applyMask ratio (SIMD on/off) | 0.617x (38.3% faster) |
| Image modifyAlpha (SIMD off) | 162.000 ms |
| Image modifyAlpha (SIMD on) | 79.000 ms |
| Image modifyAlpha ratio (SIMD on/off) | 0.488x (51.2% faster) |
| Image modifyAlpha removeColor (SIMD off) | 173.000 ms |
| Image modifyAlpha removeColor (SIMD on) | 105.000 ms |
| Image modifyAlpha removeColor ratio (SIMD on/off) | 0.607x (39.3% faster) |
| Image PNG encode (SIMD off) | 1229.000 ms |
| Image PNG encode (SIMD on) | 1074.000 ms |
| Image PNG encode ratio (SIMD on/off) | 0.874x (12.6% faster) |
| Image JPEG encode | 654.000 ms |

PR #4808, opened 2026-04-24
Another attempt to fix iPad restart Form size issue

shai-almog:
Fixed #4767

PR #4853, opened 2026-05-03
iOS: orientation-validate canvas size on iPad foreground (#4767)

### LINKED ISSUE #4767: Ipad frame size after restart

shai-almog:
## Summary

- Restores the next-runloop deferral around `updateCanvas:` in `cn1ApplicationWillEnterForeground` so UIKit has a tick to settle view bounds after the snapshot rotation.
- Cross-checks `view.bounds` against the windowScene's `interfaceOrientation` inside `updateCanvas:` and swaps the dimensions when they contradict it -- catches the failure even when the deferred call still fires before UIKit has restored bounds.

## Why

Previous attempt (#4808 / a09b9162) correctly filters `viewWillTransitionToSize:` during the background snapshot pass, but it then calls `updateCanvas:` synchronously inside `cn1ApplicationWillEnterForeground`. On iPad with UIScene, `view.bounds` is still in the snapshot orientation at that point, so the synchronous call republishes the swapped dimensions through `screenSizeChanged` -- the exact transient wrong size the user reports between stop and start in #4767.

The interface orientation reported by the windowScene is authoritative for what the user actually sees, so it can be used as a tiebreaker when the bounds and the orientation disagree.

## Test plan

- [ ] Verify on physical iPad in landscape: home → reopen no longer reports a portrait-orientation `screenSizeChanged` between stop and start.
- [ ] Verify on physical iPad in portrait: home → reopen no longer reports a landscape-orientation `screenSizeChanged` between stop and start.
- [ ] Confirm normal interactive rotation (`viewWillTransitionToSize:`) still propagates correctly.
- [ ] Confirm iPhone behavior is unchanged.
- [ ] Confirm UIScene-disabled (`ios.uiscene=false`) builds still behave correctly.

Fixes #4767

### DIFF
--- a/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
+++ b/Ports/iOSPort/nativeSources/CodenameOne_GLAppDelegate.m
@@ -248,7 +248,16 @@ static void installSignalHandlers() {
     com_codename1_impl_ios_IOSImplementation_applicationWillEnterForeground__(CN1_THREAD_GET_STATE_PASS_SINGLE_ARG);
     CodenameOne_GLViewController* vc = [CodenameOne_GLViewController instance];
     if (vc != nil) {
-        [vc updateCanvas:YES];
+        // Defer to the next runloop so UIKit can settle the view bounds
+        // after the snapshot rotation. updateCanvas itself also
+        // orientation-validates the bounds for an extra safety net under
+        // UIScene on iPad (issue #4767).
+        dispatch_async(dispatch_get_main_queue(), ^{
+            CodenameOne_GLViewController* deferredVc = [CodenameOne_GLViewController instance];
+            if (deferredVc != nil) {
+                [deferredVc updateCanvas:YES];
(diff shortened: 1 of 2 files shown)


shai-almog:
Compared 79 screenshots: 79 matched.

### Benchmark Results

- **Compilation Time:** 526 seconds

#### Build and Run Timing
| Metric | Duration |
| --- | --- |
| Simulator Boot | 164000 ms |
| Simulator Boot (Run) | 3000 ms |
| App Install | 30000 ms |
| App Launch | 31000 ms |
| Test Execution | 309000 ms |

PR #4859, opened 2026-05-03
iOS: opt-in iosReturnExitsEditing for multi-line TextArea (#4854)

### LINKED ISSUE #4854: Make TextField behavior closer to iPhone native by allowing Return to exit editing instead of inserting a newline
I hesitated about creating a bug or RFE, but settled for bug because I understand the CN1 intention is to be at par with native iOS :-).

The editing on iPhone is (IMHO) pretty well designed as illustrated e.g. by the Reminders app:

1) When you edit the task text it starts with one line and extends/wraps if you enter more text. The keyboard only has a Return button which will stop editing (not insert a newline). So this is a single-line text field that will automatically extend the number of rows, but which does not allow newlines in the text.

2) When you edit the Notes of a reminder, it works almost the same way, except now Return inserts a newline and to leave the text field you need to click another field or finish the editing by clicking the blue checkmark. So this is a multiline field, starting with a size of . What is probably missing from a UI perspective is the [Done] button but I'm guessing Apple left it out because it is quite obvious what you can do (save the task or tap another field to edit it). 

In CN1 I have not managed to replicate the  behavior in 1) with `TextField`, even with different combinations of `setRows(), setGrowByContent(), setSingleLineTextArea(), putClientProperty("iosHideToolbar", Boolean.TRUE)`. If I set ﻿`setSingleLineTextArea(true)` combined with `setGrowByContent(true)` you cannot enter a newline, but the text scrolls sideways instead of wrapping text over multiple lines. If I set `setSingleLineTextArea(false) `then tapping Return always inserts a newline, no matter how the other settings are set. 

Unless I’ve just not figured out the right combination of settings, it seems CN1 TextField misses the possibility to have growByContent but instead of having Return insert a newline it should act as Done (exiting the edit field). Maybe something like `TextField.setReturnKeyExitsMultilineEditing(boolean)` (which I guess would be true for SingleLineTextAreas)? And it should wrap long lines across multiple rows (unlike `setSingleLineTextArea(true)`).  

In addition, it would be nice to be able to change the text on the Return button to something like "Done" but is not essential (and I wonder if Apple left it out to avoid confusion between indicating that it is the editing of the text field is done or the editing of the entire task).

shai-almog:
## Summary
- Adds a guarded client property `iosReturnExitsEditing` that makes the iOS keyboard's Return key act as Done on a multi-line `TextArea` -- exits editing (firing the Done listener) instead of inserting a newline. Mirrors the iOS Reminders task-title field.
- While the flag is set, the keyboard's Return key is relabeled to **Done** (`UIReturnKeyDone`).
- Default behavior is unchanged: the flag defaults to off, only takes effect on multi-line `TextArea`s, and only intercepts an exact `"\n"` replacement so pasted multi-line text is unaffected.

Closes #4854.

## Why this shape
The reporter wants the Reminders-style growing text field where Return finishes editing rather than inserting a newline. Native iOS doesn't expose this as a built-in primitive -- Reminders implements it on a `UITextView` whose delegate intercepts `\n` in `shouldChangeTextInRange:`. We replicate that exactly, gated behind an opt-in client property so no existing layout changes behavior.

## Touchpoints
- `IOSNative.java` / `IOSImplementation.java` -- add `boolean returnExitsEditing` to `editStringAt`, populated from `cmp.getClientProperty(\"iosReturnExitsEditing\")` and forced false on single-line text fields.
- `IOSNative.m` -- bridge function gets the new `JAVA_BOOLEAN` parameter; mangled name updated.
- `CodenameOne_GLViewController.m` -- new global `BOOL currentlyReturnExitsEditing` (mirrors the existing `currentlyEditingMaxLength` pattern); set in `editStringAtImpl`; in the `UITextView` branch, forces `returnKeyType = UIReturnKeyDone` when the flag is set.
- `EAGLView.m` / `METALView.m` -- `textView:shouldChangeTextInRange:replacementText:` intercepts a single `\n` replacement when the flag is set, calls existing `keyboardDoneClicked` (which fires `stringEdit(YES, ...)` -> Java `fireDoneEvent`), and returns NO to suppress the newline insertion.

## Usage
```java
TextArea ta = new TextArea(\"\", 3, 30);
ta.putClientProperty(\"iosReturnExitsEditing\", Boolean.TRUE);
ta.setDoneListener(e -> { /* Return / Done was tapped */ });
```

## Test plan
- [ ] Build iOS port: `./scripts/build-ios-port.sh -DskipTests`
- [ ] Sample app on device: multi-line `TextArea` with `iosReturnExitsEditing=true` -- pressing Return ends editing and the Done listener fires; long input wraps to multiple visible rows.
- [ ] Same `TextArea` without the property -- Return still inserts `\n` (unchanged).
- [ ] Single-line `TextField` regardless of property -- Return still resigns first responder (unchanged).
- [ ] Pasting multi-line text into a `TextArea` with the property set -- newlines are preserved (interception is single-`\n`-only).
- [ ] Toolbar Done/Next buttons (when `iosHideToolbar` not set) still work as before.

### DIFF
--- a/Ports/iOSPort/nativeSources/CodenameOne_GLViewController.m
+++ b/Ports/iOSPort/nativeSources/CodenameOne_GLViewController.m
@@ -226,6 +226,7 @@ static CN1StatusBarTapProxyView *cn1StatusBarTapProxy = nil;
 int orientationLock = 0;
 int upsideDownMultiplier = -1;
 int currentlyEditingMaxLength;
+BOOL currentlyReturnExitsEditing = NO;
 
 #ifndef CN1_USE_ARC
 NSAutoreleasePool *globalCodenameOnePool;
@@ -357,13 +358,17 @@ void cn1_setStyleDoneButton(CN1_THREAD_STATE_MULTI_ARG UIBarButtonItem* btn) {
 void Java_com_codename1_impl_ios_IOSImplementation_editStringAtImpl
 (CN1_THREAD_STATE_MULTI_ARG int x, int y, int w, int h, void* font, int isSingleLine, int rows, int maxSize,
  int constraint, const char* str, int len, BOOL forceSlideUp,
- int color, JAVA_LONG imagePeer, int padTop, int padBottom, int padLeft, int padRight, NSString* hintString, int hintColor, BOOL showToolbar, BOOL blockCopyPaste, int alignment, int verticalAlignment) {
+ int color, JAVA_LONG imagePeer, int padTop, int padBottom, int padLeft, int padRight, NSString* hintString, int hintColor, BOOL showToolbar, BOOL blockCopyPaste, int alignment, int verticalAlignment, BOOL returnExitsEditing) {
     // don't show toolbar in iOS 8 in landscape since there is just no room for that...
     if(isIOS8() && displayHeight < displayWidth) {
         showToolbar = NO;
     }
     //CN1Log(@"Java_com_codename1_impl_ios_IOSImplementation_editStringAtImpl");
     currentlyEditingMaxLength = maxSize;
+    // Honored by the UITextView shouldChangeTextInRange: delegate (EAGLView/METALView) to
+    // intercept Return on multi-line text areas when the iosReturnExitsEditing client
+    // property is set on the editing component.
+    currentlyReturnExitsEditing = returnExitsEditing && !isSingleLine;
     dispatch_sync(dispatch_get_main_queue(), ^{
         if(editingComponent != nil) {
             [editingComponent resignFirstResponder];
@@ -615,6 +620,13 @@ void Java_com_codename1_impl_ios_IOSImplementation_editStringAtImpl
             }
             utv.text = [NSString stringWithUTF8String:str];
             utv.delegate = [[CodenameOne_GLViewController instance] eaglView];
+
+            // When iosReturnExitsEditing is set on a multi-line TextArea, present the
+            // Return key as "Done" -- the actual exit-on-return is enforced by the
+            // shouldChangeTextInRange: delegate, which intercepts a "\n" replacement.
+            if (currentlyReturnExitsEditing) {
+                utv.returnKeyType = UIReturnKeyDone;
+            }
             
             // Apply constraints for multiline text view
             // INITIAL_CAPS_WORD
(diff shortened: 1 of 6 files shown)


shai-almog:
Compared 86 screenshots: 86 matched.

### Benchmark Results

- **Compilation Time:** 112 seconds

#### Build and Run Timing
| Metric | Duration |
| --- | --- |
| Simulator Boot | 64000 ms |
| Simulator Boot (Run) | 0 ms |
| App Install | 13000 ms |
| App Launch | 6000 ms |
| Test Execution | 239000 ms |

| Metric | Duration |
| --- | --- |
| Base64 native encode | 1171.000 ms |
| Base64 CN1 encode | 1279.000 ms |
| Base64 encode ratio (CN1/native) | 1.092x (9.2% slower) |
| Base64 native decode | 696.000 ms |
| Base64 CN1 decode | 947.000 ms |
| Base64 decode ratio (CN1/native) | 1.361x (36.1% slower) |
| Base64 SIMD encode | 509.000 ms |
| Base64 encode ratio (SIMD/native) | 0.435x (56.5% faster) |
| Base64 encode ratio (SIMD/CN1) | 0.398x (60.2% faster) |
| Base64 SIMD decode | 411.000 ms |
| Base64 decode ratio (SIMD/native) | 0.591x (40.9% faster) |
| Base64 decode ratio (SIMD/CN1) | 0.434x (56.6% faster) |
| Image encode benchmark iterations | 100 |
| Image createMask (SIMD off) | 58.000 ms |
| Image createMask (SIMD on) | 9.000 ms |
| Image createMask ratio (SIMD on/off) | 0.155x (84.5% faster) |
| Image applyMask (SIMD off) | 132.000 ms |
| Image applyMask (SIMD on) | 60.000 ms |
| Image applyMask ratio (SIMD on/off) | 0.455x (54.5% faster) |
| Image modifyAlpha (SIMD off) | 121.000 ms |
| Image modifyAlpha (SIMD on) | 59.000 ms |
| Image modifyAlpha ratio (SIMD on/off) | 0.488x (51.2% faster) |
| Image modifyAlpha removeColor (SIMD off) | 157.000 ms |
| Image modifyAlpha removeColor (SIMD on) | 79.000 ms |
| Image modifyAlpha removeColor ratio (SIMD on/off) | 0.503x (49.7% faster) |
| Image PNG encode (SIMD off) | 934.000 ms |
| Image PNG encode (SIMD on) | 807.000 ms |
| Image PNG encode ratio (SIMD on/off) | 0.864x (13.6% faster) |
| Image JPEG encode | 456.000 ms |

PR #4938, opened 2026-05-13
Adjusting developer guide

LuccaPrado:
In those 2 commits I've changed the print screen for initializr to the current one and changed the url for initializr, from start.codenameone to codenameone.com/initializr as its a bit weird to expect one and be the other (if not agreed I can just revert this commit


shai-almog:
Thanks!

PR #4940, opened 2026-05-13
Adjusting rest connection example

LuccaPrado:
I was trying to test the Rest connection on cn1 and noticed that the example isnt valid anymore. 

So I changed to use the responseData

image


shai-almog:
Thanks!

PR #4984, opened 2026-05-19
Simulator: split Skins menu, fix Native Theme menu, add regression test

shai-almog:
## Summary

- **Skins menu**: only OTA downloads from the codenameone.com gallery live behind the new "Legacy Skins" submenu now. The framework default, Desktop / UWP Desktop pseudo-skins, and any local skin picked via Add Skin all stay at the top level. A single shared `ButtonGroup` spans both levels so the active skin shows as selected wherever it lives. Add Skin now persists the pick to the `skins` preference immediately (it relied on `loadSkinFile` on the reload path before, so the entry could vanish if the reload was dismissed).
- **Native Theme menu**: was silently broken for `iOS Modern` and `Android Material`. The ant `Ports/JavaSE/build.xml` only copied `iOS7Theme.res` and `winTheme.res` into the JavaSE jar, so a menu pick of either modern theme hit `JavaSEPort.class.getResourceAsStream("/iOSModernTheme.res") == null` and silently fell back to the embedded theme. Added the modern + Android legacy themes to the ant copy with `failonerror=false`, matching the existing maven pom pattern.
- **Regression test**: the existing JavaSE simulator screenshot suite gains four `native-theme-*` scenarios that drive the same preference + reload mechanism the menu uses and assert the simulator actually loaded the expected `.res`. A new `JavaSEPort.getCurrentSimulatorNativeTheme()` exposes the resolved override key; `SimulatorModeTestApp` reads it back, writes a `result=PASS|FAIL` line, and the verifier asserts on it. The runner now invokes `scripts/build-native-themes.sh` before the ant build so fresh checkouts have the CSS-compiled `.res` files ready.

## Test plan

- [ ] CI: `Test JavaSE simulator integration scripts` workflow runs the four new `javase-single-native-theme-*` scenarios green
- [ ] Local: open the Skins menu in the simulator — confirm default + user skins are top-level, OTA-downloaded skins are behind "Legacy Skins", Desktop / UWP Desktop show as radios
- [ ] Local: pick a `.skin` via Add Skin, then dismiss the reload prompt — verify the entry persists in the menu on next open
- [ ] Local: pick "iOS Modern (Liquid Glass)" from the Native Theme menu — confirm the simulator reloads with that theme applied
- [ ] Local: pick "Android Material" from the Native Theme menu — confirm the simulator reloads with that theme applied


shai-almog:

PR #5092, opened 2026-05-29
Defer lightweight Picker traversal editing until dismiss completes

jsfan3:
### Summary
- Defer lightweight `Picker` previous/next focus transfer until the popup dismiss animation has completed.
- Start editing the next component from the `disposeToTheBottom()` completion callback instead of immediately after requesting the dismiss.

### Problem
The lightweight `Picker` supports previous/next traversal buttons. When a traversal button is pressed, `endEditing()` currently calls `dlg.disposeToTheBottom()` and then immediately requests focus and calls `startEditingAsync()` on the next component.

`InteractionDialog#disposeToTheBottom()` is animated. On Android, `startEditingAsync()` creates and positions a native `EditText` overlay for the target `TextField`. If that happens while the picker popup is still being dismissed and the form content pane has only just been restored from the picker padding/invisible-area adjustment, the native editor can be positioned using stale geometry. In a reproduced Android case, traversing from a lightweight picker to a numeric `TextField` opened the numeric keyboard but rendered the field value outside of the field.

### Fix
For traversal commands that resolve to a next/previous component, this patch passes a completion callback to `disposeToTheBottom()` and starts editing only after the lightweight picker has finished dismissing. Cancel and Done without traversal continue to dismiss normally.

This keeps the existing traversal behavior, but avoids starting the next native edit session while the outgoing picker popup is still in its dismiss/layout transition.

### Verification
- Reproduced on Android with a lightweight string `Picker` followed by numeric `TextField`s and `picker.setTraversable(true)`.
- Before this patch: pressing the picker down-arrow to reach a numeric text field opens the keyboard and can render the field value outside the field.
- This patch changes the focus/edit timing so the next editor is created after the picker dismiss callback.
- Ran `git diff --check` on the branch.

### DIFF
--- a/CodenameOne/src/com/codename1/ui/spinner/Picker.java
+++ b/CodenameOne/src/com/codename1/ui/spinner/Picker.java
@@ -596,7 +596,6 @@ public class Picker extends Button {
                     currentSpinner = null;
                 }
                 restoreContentPane();
-                dlg.disposeToTheBottom();
                 if (command == COMMAND_CANCEL) {
                     // Roll back any setX calls made while the popup was showing
                     // (e.g. a custom "+7 days" button) so getDate() returns the
@@ -609,6 +608,7 @@ public class Picker extends Button {
                     preEditValue = null;
                     preEditDateValueExplicitlySet = false;
                     updateValue();
+                    dlg.disposeToTheBottom();
                 } else {
                     preEditValue = null;
                     preEditDateValueExplicitlySet = false;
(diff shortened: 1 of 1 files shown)


jsfan3:
@shai-almog Please wait before accepting this PR; I have a comment to make that I'll post shortly...

jsfan3:
@shai-almog Small clarification about verification status.

This fix is based on a reproduced Android issue and on code-path analysis of the lightweight `Picker` dismissal flow, but it has not yet been verified by rebuilding Codename One locally with this patch and running the patched framework on a device.

The diagnosis and patch were prepared with assistance from generative AI. The observed behavior is real and reproducible in my app: traversing from a lightweight `Picker` to a numeric `TextField` can open the Android native editor with the field value rendered outside of the field. The proposed fix is to defer `requestFocus()` / `startEditingAsync()` until the picker dismiss animation has completed, which appears consistent with the suspected race/layout timing issue.

If you think this needs direct validation before review/merge, I can test it properly by building Codename One locally with this patch and running the app against that patched build. That may take some time, but I can do it.

shai-almog:
@jsfan3 we have CI that checks most of these things and is now pretty thorough (albeit not perfect). The problem is that it's really hard to run that CI on 3rd party PRs...  Even when I accept the PR and let CI run some things fail due to the complexity of our CI scripts and security concerns.

Due to that I'd much rather get bug reports than PRs. I reviewed this and it looks safe so I'll merge it and it will run in CI post merging.

PR #5211, opened 2026-06-10
Simulator UX: restore Slow Motion, fix rotate/zoom, standalone default, menu cleanup

shai-almog:
Reworks the JavaSE simulator chrome around four user-reported issues.

## 1. Slow Motion restored
The **Slow Motion** toggle (`Motion.setSlowMotion`, 50× animation slowdown) had been commented out and orphaned. It's re-enabled under **Tools** and initializes from `Motion.isSlowMotion()`.

## 2. Rotation no longer drops the zoom/fit
Both rotate paths — the **Rotate** menu item and the app-frame `RotateAction` toolbar button — sized the canvas to `skin × zoomLevel`, discarding the fit-to-screen factor that only the Zoom toggle computed. So after rotating you had to re-toggle Zoom to get the skin to fit again. The fit logic is now extracted into shared helpers (`fitToScreenZoomFactor` / `computeSkinDisplaySize`) used by all three call sites, so rotation preserves the active fit/zoom state.

## 3. Standalone simulator is now the default
`Executor.main` no longer forces `cn1.simulator.useAppFrame=true`; it sets a `cn1.simulator.interactiveSimulator` context marker instead. `JavaSEPort.computeUseAppFrame` resolves the shell as:

1. explicit `-Dcn1.simulator.useAppFrame=true|false` always wins (so `SimulatorWindowModeVerifier` and tests still force a mode),
2. otherwise the interactive simulator honours the persisted **Single Window Mode** preference,
3. otherwise **standalone**.

Existing explicit opt-ins are preserved; the out-of-the-box default is now the plain device simulator instead of the multi-panel window.

## 4. Menu hierarchy cleaned up (Device / Simulate / Tools)
The confusing **Simulator** vs **Simulate** split is gone:

- **Simulator → Device**: the device and its window — Rotate, Zoom, Skin▸, Native Theme▸, screenshots, Single Window Mode, Always on Top, Exit. **Skins** and **Native Theme** are folded in as submenus instead of top-level menus.
- **Simulate**: only faked device state/events (Pause, Send App Argument, Location, Push, Biometric, NFC, Dark/Light, Larger Text, In-App Purchase, etc.).
- **Tools**: developer tools and diagnostics, including the restored **Slow Motion** and the relocated **Auto Update Default Bundle**.

Items are still created with their listeners as before, then re-assembled in one deliberately-ordered block, preserving the existing `appFrame == null` gating so Single Window mode keeps surfacing rotate/zoom/screenshots on its toolbar rather than in the menu.

## Testing
- `mvn -pl javase -Plocal-dev-javase compile` (JDK 8) — BUILD SUCCESS.

### DIFF
--- a/Ports/JavaSE/src/com/codename1/impl/javase/Executor.java
+++ b/Ports/JavaSE/src/com/codename1/impl/javase/Executor.java
@@ -129,8 +129,14 @@ public class Executor {
     
     
     public static void main(final Class launcherClass, final String[] argv) throws Exception {
-        if (System.getProperty("cn1.simulator.useAppFrame", null) == null) {
-            System.setProperty("cn1.simulator.useAppFrame", "true");
+        // Mark that we are launching the interactive simulator. Only in this
+        // context does the persisted "Single Window Mode" preference take
+        // effect (see JavaSEPort.computeUseAppFrame). The default shell is the
+        // standalone device simulator; Single Window mode is opt-in via the
+        // Device menu. We intentionally do NOT force cn1.simulator.useAppFrame
+        // here anymore so the standalone simulator is the out-of-the-box mode.
+        if (System.getProperty("cn1.simulator.interactiveSimulator", null) == null) {
+            System.setProperty("cn1.simulator.interactiveSimulator", "true");
         }
         if(IS_MAC) {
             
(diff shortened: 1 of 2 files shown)


shai-almog:

PR #5516, opened 2026-08-04
Fix the attach race in HealthEdtDeliveryTest.aFacadeActionDeliversOnTheEdt

shai-almog:
## The problem

`HealthEdtDeliveryTest.aFacadeActionDeliversOnTheEdt` fails **8/8 times** when run on its own, and flakes roughly **1 in 10** when the whole class runs:

```
AssertionFailedError: a result must arrive on the EDT on every backend ==> expected: <true> but was: <false>
  at HealthEdtDeliveryTest.assertDeliveredOnEdt(HealthEdtDeliveryTest.java:112)
  at HealthEdtDeliveryTest.aFacadeActionDeliversOnTheEdt(HealthEdtDeliveryTest.java:316)
```

`pr.yml` runs `core-unittests` on every PR, so this was an intermittent red build waiting to happen.

## Root cause

**The product code is correct** — `EdtResult` hops to the EDT in every failing run. The test loses a race it cannot win.

`Health.openHealthSettings()` completes its `EdtResult` *before it returns*:

```java
public AsyncResource<Boolean> openHealthSettings() {
    AsyncResource<Boolean> out = new EdtResult<Boolean>();
    out.complete(Boolean.FALSE);   // -> callSerially(Deliver)
    return out;
}
```

The test then attaches `.onResult(landing)` afterwards. By then the EDT has usually already run the queued `Deliver`, and `AsyncResource.ready` takes its already-settled branch (`AsyncResource.java:451`), invoking the listener **inline on the attaching thread**. The landing records `isEdt() == false` for a delivery that genuinely happened on the EDT.

Instrumenting the delivery proves it fires from the attach, not from the queued runnable:

```
at AsyncResource.ready(AsyncResource.java:470)   <-- runImmediately, already-done path
at AsyncResource.onResult(AsyncResource.java:634)
at HealthEdtDeliveryTest$1.run(...)
at com.codename1.ui.Display.invokeAndBlock(Display.java:1561)
```

This is the same race the class already documents on `anAggregateDeliversOnTheEdt`. The other six tests dodge it by holding the backend (`holdRead`/`holdWrite`/`holdAggregate`/`holdDelete`) until the listener is attached. The facade settles synchronously, so there is nothing to hold — which is why this one test was left exposed.

A second detail widens the window: JUnit runs these on `main`, not the EDT, so `CN.invokeAndBlock` takes its non-EDT branch (`Display.java:1561`, a plain `r.run()`). The real EDT spins idle alongside and grabs the queued delivery almost instantly. That is why isolation fails deterministically while a busier full-class run usually wins.

## The fix

Park the EDT behind a barrier queued *ahead* of the delivery, and release it once the listener is attached. Serial calls run FIFO, so the ordering is guaranteed rather than hoped for.

```java
CN.callSerially(new Runnable() {
    public void run() {
        try { attached.await(); } catch (InterruptedException ex) { Thread.currentThread().interrupt(); }
    }
});
Health.getInstance().openHealthSettings().onResult(landing);
attached.countDown();
```

Also in this change:

- **Corrected the `assertDeliveredOnEdt` docs.** They stated these tests run on the EDT and that `invokeAndBlock` keeps the loop pumping. Neither holds, and the gap between the comment and the behaviour is what made the race invisible. Replaced with what actually happens, plus the attach-before-settle rule every test here depends on.
- **Moved the facade test's javadoc back onto the facade test.** It had drifted upward and was sitting stacked above `everyPublicHealthResourceDeliversOnTheEdt`'s own javadoc, documenting a test ~100 lines away.

No production code is touched.

## Verification

| Run | Before | After |
|---|---|---|
| `-Dtest=HealthEdtDeliveryTest#aFacadeActionDeliversOnTheEdt` | 8/8 fail | **0/12 fail** |
| `-Dtest=HealthEdtDeliveryTest` (full class) | 1/10 fail | **0/30 fail** |

`mvn verify` on `core-unittests` is green: SpotBugs `BugInstance size is 0`, PMD, Checkstyle and Spotless all clean.

### DIFF
--- a/maven/core-unittests/src/test/java/com/codename1/health/HealthEdtDeliveryTest.java
+++ b/maven/core-unittests/src/test/java/com/codename1/health/HealthEdtDeliveryTest.java
@@ -35,6 +35,7 @@ import java.util.Map;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.junit.jupiter.api.Assertions.*;
@@ -330,14 +331,23 @@ class HealthEdtDeliveryTest extends UITestBase {
                 CN.callSerially(new Runnable() {
                     public void run() {
                         try {
-                            attached.await();
+                            // Bounded on purpose: a blocker that outlived its
+                            // release would park the EDT for the rest of the
+                            // suite, turning one failure into a hang.
+                            attached.await(10, TimeUnit.SECONDS);
                         } catch (InterruptedException ex) {
                             Thread.currentThread().interrupt();
                         }
                     }
                 });
-                Health.getInstance().openHealthSettings().onResult(landing);
-                attached.countDown();
+                try {
+                    Health.getInstance().openHealthSettings()
+                            .onResult(landing);
+                } finally {
+                    // In a finally so a throw above surfaces as itself rather
+                    // than as a suite-wide hang behind a still-parked EDT.
+                    attached.countDown();
+                }
             }
         }, true);
     }
--- a/maven/core-unittests/src/test/java/com/codename1/health/HealthEdtDeliveryTest.java
+++ b/maven/core-unittests/src/test/java/com/codename1/health/HealthEdtDeliveryTest.java
@@ -34,6 +34,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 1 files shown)


shai-almog:
Good catch — fixed in 165335943.

You're right that the failure mode was worse than the flake being fixed: a skipped `countDown()` would leave the EDT parked in an unbounded `await()`, so the suite hangs and the throw that caused it never gets reported.

Applied both suggestions — `countDown()` in a `finally`, and the wait bounded at 10s as a backstop for anything the `finally` can't reach:

```java
CN.callSerially(new Runnable() {
    public void run() {
        try {
            attached.await(10, TimeUnit.SECONDS);
        } catch (InterruptedException ex) {
            Thread.currentThread().interrupt();
        }
    }
});
try {
    Health.getInstance().openHealthSettings().onResult(landing);
} finally {
    attached.countDown();
}
```

Verified with a throwaway probe that replicates the barrier around a deliberate throw:

```
PROBE later serial call ran, isEdt=true
PROBE later test saw EDT alive=true
PROBE blocker released
[ERROR] ZzTmpHangProbeTest.throwingInsideTheBarrier... » IllegalState simulated facade failure
Tests run: 2, Failures: 0, Errors: 1 -- Time elapsed: 0.106 s
```

The `IllegalStateException` surfaces as itself, the blocker is released, and a later test's serial call still runs on a live EDT. Elapsed 0.106s, so the `finally` did the releasing — the 10s timeout was never reached. Probe deleted after the run.

Re-verified after the change: 0/10 failures running the method alone, 0/25 for the full class, and `mvn verify` still green (SpotBugs `BugInstance size is 0`, PMD/Checkstyle/Spotless clean).

PR #5629, opened 2026-08-30
JS port: apply the text layer's DOM changes with the frame they belong to

shai-almog:
## The defect

`f691c7e4a6` (#5552) gave the port a DOM text layer, which makes every frame two
halves: the pixels the canvas draws and the elements the layer positions. Only
one of them is issued when the frame is decided.

`promote()`, `endComponent()` and `syncToForm()` wrote to the document **during
the component paint**. The canvas commands that same paint records are not
replayed and shipped until the frame is drained an animation frame later. The
browser composites in between, so it renders one half of the new frame over the
other half of the old one:

- text released early leaves a hole where the glyphs still are on screen;
- text added early is drawn twice, once in each place.

Nothing about the layer's bookkeeping was wrong, which is why it survived
review. It was applied at the wrong moment.

Reproduced on the BuildCloud console (7.0.267, real Chrome, 30fps video): a
click on a navigation row blanks **every label and every icon on the form** for
a frame, and draws the outgoing and incoming label of that row superimposed.
Icons go through `drawString` too, so they are runs and blank with the text.

## The fix

Every mutation becomes a `TextLayerOp` (`ExecutableOp`) recorded into the same
buffer as the draws around it, replayed into the same `SurfaceCommandRecorder`
(`OP_TEXT_*`, 90-96), and shipped in the same flush message. The host applies it
in the task that replays that frame's canvas commands, in the draw order the
paint gave it, so no rendering opportunity exists between the two halves.
`OP_BLUR_SELF_REGION` already rides the stream for the same reason.

`syncToForm` moves from the drain to `flushGraphics`. By drain time the frame's
ops are already snapshotted, so a release recorded there shipped with the *next*
frame — which is what put a removed component's text over the pixels that
replaced it. In `flushGraphics` the components have finished painting and the
buffer is still open.

Two things fall out: the retained `CSSStyleDeclaration` handles are gone (style
is written by opcode now), and with them the two `getStyle()` property reads per
run the class had to cache to stay off the barrier.

## Measurement

Scripted 1028-frame interaction, real Chrome, 30fps, counting **transient**
brightness spikes (a frame brighter than the one before that recovers within
three — a navigation is a step that stays, and is not counted):

| console build | transient blank flashes |
| --- | --- |
| 7.0.267 as released | 6 |
| this branch | 0 |

0 is what the same run measures with the layer disabled entirely
(`?cn1TextLayer=0`), so the flicker is gone rather than merely smaller.

[...]

PR #5552, opened 2026-08-16
JavaScript port: real DOM text, incremental semantics, and native-resolution rendering

shai-almog:
## What this does

Makes the JavaScript port feel like a web page without replacing the renderer.

The obvious approach -- emit DOM per component -- is the one Flutter and Uno both tried and reversed. What they kept is the useful half: canvas for pixels, a targeted DOM layer for text, semantics and input. Flutter's canvas apps ship `<flt-semantics>` for ARIA and `<flt-text-editing-host>` for a real positioned `<input>`. This PR takes the port to that shape.

## The layers

Two DOM layers sit above the canvas, which is already `role=presentation` / `aria-hidden`:

- **`#cn1-text-layer`** carries the visible text. `BufferedGraphics.drawString` hands each run to the layer instead of the canvas, so what the user sees is real text -- selectable, findable, rasterized by the browser. Codename One stays the sole layout authority: a run arrives already broken and placed, so it is one `white-space:pre` element at an absolute coordinate and the browser cannot wrap or reflow it. Text measurement is untouched and stays on the worker's `OffscreenCanvas`.
- **`#cn1-accessibility-tree`** carries the ARIA projection, now updated incrementally.

Text that stays on the canvas by design: offscreen targets (transition buffers, `paintLock`, `ComponentImage`, `screenshot` -- a structural gate, since those use plain `HTML5Graphics`), cell renderers (one instance stamped at N positions cannot key a pooled element), anything outside the displayed form (a modal dialog paints the form beneath it as its backdrop), shape clips and non-identity transforms. Bitmap fonts need no exclusion -- `Graphics.drawString` renders a `CustomFont` itself and never reaches the implementation.

## Bugs fixed along the way

**The overlay was rebuilt on every invalidation.** `accessibilityTreeChanged` did `setInnerHTML("")` and recreated every element with fresh listeners. `CHANGE_BOUNDS` is raised by every `setX/setY/setWidth/setHeight`, so this ran on every scroll step and discarded DOM focus and any in-progress text selection each time. Nothing selectable could live there until it diffed.

**The port rendered at 1x on every HiDPI display.** `port.js` pinned `devicePixelRatio` to 1 and called it working "in CSS pixels end-to-end". Codename One addresses device pixels -- the iOS port detects the retina factor and scales the values it hands the native primitives -- so this was a misreading of the API, and it meant the browser upscaled a 1x bitmap on every retina screen. Text showed it first. Now 750x1334 backing behind a 375x667 CSS box at dpr 2.

[...]

PR #5634, opened 2026-08-30
JS port: device pixel ratio for breakpoints, and a millimetre that means the same thing at every ratio

shai-almog:
Two consequences of #5552's native-resolution change that were left behind, both
found while chasing a broken responsive layout in the BuildCloud console.

Independent of #5629 — different files, different failure, no ordering between them.

## 1. Application code cannot write a width breakpoint any more

`getDisplayWidth()` reports device pixels, and since 7.0.267 this port reports
them too rather than CSS pixels. A threshold written in CSS pixels therefore
moves with the display: a 390pt phone at ratio 3 reports **1170**, so *"is this
narrower than 600"* answers **no** on a phone. The console lost its phone layout
exactly that way, and could no longer be made to appear in a browser's
responsive mode — which keeps a desktop User-Agent, so a UA check does not save
you, and applies a pixel ratio.

Nothing already reachable answers the question. `getDeviceDensity()` buckets the
ratio, so anything derived from it steps rather than scales. The ratio itself is
what is needed and the browser knows it exactly, so this exposes it beside the
other `browser.window.*` properties.

## 2. A millimetre was not the same size at every ratio

`convertToPixels()` read a nominal dpi out of one of nine Android-shaped density
buckets. That only lands on Codename One's 160dpi-per-CSS-pixel baseline at
ratios 1 and 2:

| ratio | bucket | effective dpi | |
| --- | --- | --- | --- |
| 1 | MEDIUM 160 | 160 | ok |
| **1.5** | MEDIUM 160 | 107 | **33% too small** |
| 2 | VERY_HIGH 320 | 160 | ok |
| **2.5** | HD 540 | 216 | **35% too large** |
| **3** | HD 540 | 180 | **12.5% too large** |

Ratio 1.5 is ordinary — plenty of Android hardware, and Windows at 150% scaling.
Ratio 3 is every recent iPhone.

Measured on the console at 390pt, text rendered **13.9–25.8 CSS px at ratios 1
and 2** and **15.6–29.1 at ratio 3**. After the change the same screen measures
13.9–25.8 at ratios 1, 1.5, 2, 2.5 and 3 alike. Verified in both Chromium and
Firefox.

`getDeviceDensity()` keeps its buckets deliberately: that one picks which
resolution of an image to load, a real choice between a handful of assets that
wants to be quantised. This is a continuous conversion and does not. An explicit
`?density=` override is still honoured exactly as given.

## Blast radius — please read before merging

The second commit changes rendering for **every JavaScript-port app at any ratio
other than 1 or 2**. At 3x everything gets 12.5% smaller; at 1.5x a third
larger. That is the correction, but it is not a quiet one, and it will move
screenshot goldens for any app not captured at ratio 1.

The first commit adds a property and changes no behaviour.

## Known gap, not addressed here

`refreshDevicePixelRatio()` is wired to the resize event, on the reasoning that
resize is when the ratio can change. A ratio can change without one — a
browser's responsive mode switching device profile is the case that prompted
this. A `matchMedia('(resolution: Ndppx)')` listener would catch it. Left alone
because it is a separate change with its own risk.

### DIFF
--- a/Ports/JavaScriptPort/src/main/java/com/codename1/impl/html5/HTML5Implementation.java
+++ b/Ports/JavaScriptPort/src/main/java/com/codename1/impl/html5/HTML5Implementation.java
@@ -4605,9 +4605,42 @@ public class HTML5Implementation extends CodenameOneImplementation {
     @Override
     public int convertToPixels(int dipCount, boolean horizontal) {
         if (ppi == 0) {
-            ppi = JavaScriptDisplayMetrics.pixelsPerMillimeter(getDeviceDensity());
+            ppi = millimetreScale();
         }
         return (int) Math.round(((float) dipCount) * ppi);
+    }
+
+    /**
+     * Device pixels per millimetre, taken from the display's actual pixel ratio.
+     *
+     * <p>Deliberately not {@code pixelsPerMillimeter(getDeviceDensity())}. That table sorts a
+     * display into one of nine Android-shaped buckets and reads a nominal dpi out of it, which
+     * only lands on Codename One's own baseline of 160dpi-per-CSS-pixel at ratios 1 and 2.
+     * Everywhere else it misses, and by a lot: ratio 3 falls in the HD bucket and is rendered
+     * 12.5% too large, ratio 2.5 lands in the same bucket and is 35% too large, and ratio 1.5 --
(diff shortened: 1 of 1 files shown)


shai-almog:
Compared 181 screenshots: 181 matched.
✅ JavaScript-port screenshot tests passed.

PR #5815, opened 2026-09-13
Certificate wizard: register Macs, report the right slot, check the profile signs this app

shai-almog:
Fixes #5773 and the client half of #5793. Four independent defects plus a preflight gap.

## 1. A Mac cannot be registered (#5773)

`CloudSigningService.registerDevice` hardcoded `"IOS"`. The signing service has always accepted `MAC_OS` — only the dialog never asked. So the wizard offered a `MAC_APP_DEVELOPMENT` profile type that nothing could satisfy: `isUsableDevice` correctly hides an iPhone from a Mac profile's picker, and no Mac could be registered to put in it. A profile type you can select and never create.

The registration dialog now picks the platform, and the UDID hint follows it — a Mac is registered by its hardware UUID, not by an iPhone's 25-character identifier.

The reporter's screenshot is the *iOS* profile device picker, where hiding Macs is correct. The bug is the missing registration path, not that screen.

## 2. The overview announced readiness it did not have (#5773)

Both cards read row zero of an unfiltered list:

```java
state.certificates.isEmpty() ? "-" : state.certificates.get(0).displayName()
state.profiles.isEmpty()     ? "-" : state.profiles.get(0).name()
```

So an account holding a Mac certificate and a development profile was told **"Apple distribution certificate: Ready — dtest11 MAC APP DISTRIBUTION"** and **"App Store profile: Ready — dtest11 Development"**. Neither slot was filled.

To be fair to the code: the wizard was never going to *use* either for an iOS build. `WizardDecisions.certificateTypeSatisfies` has always been strict and needed no change. But that panel is why the reporter concluded it was signing iOS builds with Mac App Store assets, and a readiness panel that reports readiness it does not have is worse than no panel at all.

The selection moved to `WizardDecisions`, where it is testable without a display. The profile card now requires the type *and* that Apple has not marked the profile INVALID — "Ready" for a profile the next build cannot sign with is the same false assurance.

## 3. A downloaded "Apple Development" certificate went to the release slot

`afterCertificateDownload` tested `"IOS_DEVELOPMENT".equals(...)` alone, while every other decision in the class goes through `isDevelopmentCertificate`, which also counts Apple's generic `DEVELOPMENT` type. A manual download of an *Apple Development* certificate therefore offered to install into `codename1.ios.release.certificate`, overwriting the unqualified key with it.

## 4. "Sync with Apple" calls both halves (#5793)

[...]

Issue #147, opened 2015-03-27
Orientation changes look wrong on iOS


ghost:
Original [issue 147](https://code.google.com/p/codenameone/issues/detail?id=147) created by codenameone on 2012-04-13T19:21:06.000Z:

<b>What steps will reproduce the problem?</b>
1. Rotate your device
2. Open your eyes ;-)

<b>What is the expected output? What do you see instead?</b>
It looks like the screen is rotated first and then resized. The same operation on a native app look like both operations are done at the same time.

<b>What version of the product are you using? On what operating system?</b>
Did the test on the Kitchen Demo from trunk on April 12th on an iPad 2.

shai-almog:
Closing per the issue-tracker triage (see [`ISSUE_TRIAGE_REPORT.md`](https://github.com/codenameone/CodenameOne/blob/master/ISSUE_TRIAGE_REPORT.md) + [`STALE_REVIEW.md`](https://github.com/codenameone/CodenameOne/blob/master/STALE_REVIEW.md)).

Filed in 2012 with "rotate, open your eyes" as the only repro. The iOS rotation code has since been rewritten to use [`viewWillTransitionToSize:withTransitionCoordinator:`](https://github.com/codenameone/CodenameOne/blob/master/Ports/iOSPort/nativeSources/CodenameOne_GLViewController.m#L3486) (the iOS 8+ API that fixes the rotate-then-resize behavior the report describes). No repro and no testable signal against the current code.

PR #5874, opened 2026-09-20
Certificate wizard: a failed profile sync is a failed sync, 404 included

### LINKED ISSUE #5832: [Bug] Sync with apple doesn't
### Summary

Image

I deleted these three keys using apple's interface about 20 minutes ago.  they no longer exist.  This
is a new, separate run of the certificate wizard.  It started empty and I did a sync with apple.
So where did these three keys come from?

### Minimal reproducer

delete a profile on developer.apple.com, then run the certificate wizard and sync.

### Steps to reproduce

like I said

### Expected and observed behavior

shouldn't recover deleted content from ?? 

### Codename One version

updated against master

### JDK version

n/a

### Target environment

eclipse

### Last known working version

_No response_

### Logs and diagnostics

```shell

```

### Screenshots or recordings

_No response_

### Report checks

- [x] I searched existing issues and discussions for this behavior.
- [x] The reproducer contains no credentials, signing keys, or private customer data.
- [x] I understand that triage is not a commitment to a fix date.

shai-almog:
Fixes #5832. Paired with codenameone/BuildCloud#153, which is the server half.

## What #5832 actually was

The reporter deleted three provisioning profiles in the Apple portal, ran "Sync with Apple", was told it worked, and watched all three stay on screen.

The server-side profile reconcile had shipped that same morning (BuildCloud #147, deployed ~02:10 UTC; the report is 18:30 UTC). What had not shipped was **this** half. `reconcileProfiles` entered `CloudSigningService` in #5815 and first reached a release in **7.0.272 (09-18)**, four days later - checked against the published artifacts rather than master:

```
$ javap -v cn1-certificatewizard-common-7.0.271 .../CloudSigningService.class | grep -i reconcile
  ... CertificatesApi.reconcileCertificates ...        # and nothing else
$ javap -v cn1-certificatewizard-common-7.0.272 .../CloudSigningService.class | grep -i reconcile
  ... CertificatesApi.reconcileCertificates ...
  ... ProfilesApi.reconcileProfiles ...
```

His sync had literally never asked about profiles, so nothing about profiles could change. **That is fixed for anyone on 7.0.272**, which is why this PR is not about it.

## What this fixes

The way the same report can arrive again from a wizard that *does* call the route: the profile half was allowed to answer **404 or 405 and still be reported as a successful sync**.

```java
if (ok(rr) || endpointAbsent(rr)) {   // 404 || 405
    callback.completed(Result.<Void>ok(null));
```

The shim was for "a wizard newer than the signing service it is talking to". That window never existed in a release - the service route shipped first, and there is only ever one deployed service - and it cost more than it could have bought, because **404 is also what the service answers when Apple said 404**. `AppleErrorTranslator`'s message for that case is, verbatim:

> Apple no longer has this item - it was probably deleted in the Apple Developer portal. Use "Sync with Apple" to bring your Codename One account back in step.

The one reply that most needed showing was the one being reported as success.

This also lands the other side of the BuildCloud change: an inventory Apple's reply could not settle now answers 502 instead of 200. Both reconcile routes already declare 502 in the `specs/openapi.json` this client is generated from, and neither declares 404 - so no spec change, and the swallow had no contract behind it either.

## Testing

There was no way to test any of this: the class builds its API clients in its constructor via `RestClients.create`, which wants a live CN1 runtime, and `Response` has no public constructor. So the sequencing moves to a package-private overload taking the two clients, and the fixture that makes a `Response` sits in `com.codename1.io.rest` under `src/test`.

```
mvn test -pl common -Dcn1.certificatewizard.skipTests=false   # 88 tests, 0 failures
```

The switch matters: the module skips its tests by default, and without it the build reports success having run nothing.

A/B probed by restoring the shim - exactly `aProfileHalfThatAnsweredNotFoundIsAFailedSync` and `...MethodNotAllowed...` go red. The probe was re-run after confirming the edit had applied, because the first attempt silently did not and reported all six passing.


shai-almog:
@codex review

